PR #17, opened 2017-12-02
Set range limit on mouse_x and mouse_y values

### LINKED ISSUE #16: Mouse_x and mouse_y returning negative values
**Test script:**

```python
from p5 import *
def setup():
    size(640, 360)
    background(255)

def draw():
    print(mouse_x, mouse_y)

run()
```
**Expected behavior:**
mouse_x and mouse_y respectively return values in the intervals [0, width] and [0, height]

**Actual behavior**
When quickly moving mouse outside of the output screen, negative values, and values exceeding width and height, get returned by mouse_x and mouse_y.

**p5 version:**
p5-0.3.0a2
**Python version:**
3.6.3
**Operating System:**
Mac OS 10.11.6 (El Capitan)

parsoyaarihant:
Fixes #16 .

Pyglet returns cursor values which my not be in the range of scren size. Hence I have constraint the cursor values to the size of sketch

### DIFF
--- a/p5/sketch/events.py
+++ b/p5/sketch/events.py
@@ -177,8 +177,8 @@ class MouseEvent(Event):
     """
     def __init__(self, action, x, y, change=(0, 0), scroll=(0, 0),
                  buttons=None, modifiers=0, handler_name=None):
-        self.x = x
-        self.y = builtins.height - y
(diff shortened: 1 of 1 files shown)


Manindra29:
This looks good to me.
cc: @abhikpal

Issue #26, opened 2018-01-04
[Feature Request] Vsync option available before/during initial call to run()


lajaybird:
As brought up in a previous issue I had posted (at [p5/issues/25](https://github.com/p5py/p5/issues/25)), I was having an issue with flickering and graphical errors on my laptop. I discovered that the solution was to turn on Vsync, but this can only be done by editing the module and cannot be done in userspace.

innovationchef:
I tried this with the same `shape_primitives.py ` file and realized that flickering occurs when vsync is assigned True and behave as expected when assigned with None or False. Could you please describe the enhancement you are expecting?

abhikpal:
@lajaybird, 

> I discovered that the solution was to turn on Vsync, but this can only be done by editing the module and cannot be done in userspace.

Ideally we would want something like to be an internal setting that we don't expose to the userspace. Ideally, sketches shouldn't flicker at all.


Hey @innovationchef 

> I tried this with the same shape_primitives.py file and realized that flickering occurs when vsync is assigned True and behave as expected when assigned with None or False. Could you please describe the enhancement you are expecting?

As I just mentioned, the goal is to not have _any_ flickering. 

> vsync is kept False in the current version of p5 and the script runs perfectly without flickering issues on my system.

That's the thing, we want a fix that works on all machines and this configuration should happen internally and not from the userspace.

innovationchef:
I am not very sure about this. Just a thought:
Could it be something related to preparing the next frame before and outside the update(dt) function? Something similar to preloading the image in opengl instead of calling it in update(dt) function (refactoring)?

PR #27, opened 2018-01-04
Resolutions for issues #26 and #16

lajaybird:
Issue #26: vsync was an issue with my own usage of p5 and the solution was only obtainable through edits in the module itself in p5/sketch/base.py.
The function vsync() is very similar in implementation to size() and can be called in the user's defined setup() function.

It should also be noted that this change also includes replacing the default vsync setting to True. This can be modified if need be.

### DIFF
--- a/p5/pmath/vector.py
+++ b/p5/pmath/vector.py
@@ -408,7 +408,7 @@ class Vector(Point):
     def random_2D(cls):
         """Return a random 2D unit vector.
         """
-        x, y = 2 * (np.random.random(2) - 0.5)
+        x, y = np.random.random(2)
         vec = cls(x, y)
         vec.normalize()
         return vec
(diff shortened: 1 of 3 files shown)


abhikpal:
Amazing! This seems to be a very reasonable addition. 

>  The function vsync() is very similar in implementation to size() and can be called in the user's defined setup() function.

Yep. That makes perfect sense. We also don't want the user to call vsync multiple times. We could of course raise errors when `vsync` is called from outside `setup()`. Alternatively: instead of making `vsync()` an independent function we could add an extra key word argument to `run()`.

I am leaning more on making vsync a kwarg to ```run()```, unless, of course, there is a stronger argument against doing that.

lajaybird:
I wondered about adding a kwarg to `run()` as well and I think it would make an equally fine solution. I couldn't make an argument against either way as of now.

PR #29, opened 2018-01-04
Fixed issue #16: negative mouse_x and mouse_y values.

lajaybird:
The issue with negative values of the mouse position was the result of the user moving the mouse past the left and top edges of the sketch during runtime.
The addition of a simple call to `max()` helped resolve the problem.


lajaybird:
Pardon me, I didn't notice this was addressed already [here](https://github.com/p5py/p5/pull/17).

Issue #31, opened 2018-01-06
Implement saveFrame()


friedkeenan:
I tried to save a frame but saw that it wasn't in this library, so I looked up how to do it in pyglet:

    pyglet.image.get_buffer_manager().get_color_buffer().save(filename)

You could then make a save_frame(filename) function that maybe parses the string like processing does to replace the "###" with the right number, and then it runs that line of code.

It's not urgent or anything, but I think it'd be nice to have

abhikpal:
Hey, sorry for getting back to this so late.

@innovationchef used your suggestion here to to implement this in #33, it's worth looking at @Manindra29 and my comment on the thread there if you're interested in implementing this.

parsoyaarihant:
`saveFrame` is implemented in p5py: https://github.com/p5py/p5/commit/1885fda916dc493f8efa3b1cc0bdc58ebf894ce2

PR #33, opened 2018-02-17
Integrated saveframes functionality

innovationchef:
Hey @Manindra29 ,
I have implemented the saveframe function as discussed in issue #31. 
I updated the argument of the run() function from -
`run(sketch_setup=None, sketch_draw=None, frame_rate=60)`
to 
`run(sketch_setup=None, sketch_draw=None, frame_rate=60, save = False)`
This will save the frames with file names 0 to n; where n is dependent on frame_rate.

We can merge the saved frames to create gifs, videos etc. Any suggestions?
Please let me know if I should change the framework of saveframe implementation.

### DIFF
--- a/p5/sketch/base.py
+++ b/p5/sketch/base.py
@@ -230,6 +230,8 @@ def update(dt):
                     window.set_visible(True)
             else:
                 draw()
+                if saveframes==True:
+                    pyglet.image.get_buffer_manager().get_color_buffer().save(filename= str(builtins.frame_count) +'.png')
                 redraw = False
         for function, event in handler_queue:
             function(event)
@@ -251,7 +253,7 @@ def fix_handler_interface(func):
     else:
         return func
 
-def run(sketch_setup=None, sketch_draw=None, frame_rate=60):
+def run(sketch_setup=None, sketch_draw=None, frame_rate=60, save = False):
     """Run a sketch.
 
     if no `sketch_setup` and `sketch_draw` are specified, p5 automatically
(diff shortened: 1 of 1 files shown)


Manindra29:
Hey,
We'd like saveFrame to match as much as possible with the processing api's [saveFrame()](https://processing.org/reference/saveFrame_.html). Can you write this as a function instead of a param for the run method?

abhikpal:
> We'd like saveFrame to match as much as possible with the processing api's saveFrame(). Can you write this as a function instead of a param for the run method?

+1 for this, actually. It's also rarely the case that someone needs a global `save_frame` function.

innovationchef:
Yeah, I will fix it. Thanks, @Manindra29  for updating the project details. Now, I have got a clear idea of how p5py is expected to be built.

innovationchef:
I tried to keep it same as given in the docs. Could you please provide me with a link to SaveFrame() Java implementation Code?

abhikpal:
> I tried to keep it same as given in the docs. Could you please provide me with a link to SaveFrame() Java implementation Code?

Judging by a simple grep on the entire Processing code, this seems to be a good place to start: https://github.com/processing/processing/blob/master/core/src/processing/core/PApplet.java

I haven't looked at the code in ages, but if I remember correctly, you will have to look around a bit before you get to the _actual_ saveFrame() implementation.

Also, minor note: I haven't looked at the code in details, but could we maybe rename the method to ```save_frame``` to be consistent with the Python conventions?

abhikpal:
Closing this as we've moved the backend to vispy and this PR relies on pyglet. Feel free to open if you reimplement this in vispy.

PR #36, opened 2018-03-09
Fixed distance() of class Vector by adding np.sqrt

Devrim-Celik:
Since the idea of the distance method is to (probably) implement the euclidean distance, we need to calculate the square-root (which is missing in the current version) of the squares of differences.

### DIFF
--- a/p5/pmath/vector.py
+++ b/p5/pmath/vector.py
@@ -93,7 +93,7 @@ class Vector(Point):
         :rtype: float
 
         """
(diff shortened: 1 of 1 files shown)


abhikpal:
Brilliant! Shouldn't you be using ```np.sqrt```? I've merged your change into ```develop``` with ```np.sqrt``` instead of ```np.square```

SneakySheep:
yes of cause, thx :)

PR #38, opened 2018-03-09
raise ValueError when Magnitude of Vector is 0

### LINKED ISSUE #37: raise an error, when Vector.normalize() tries to normalize a "zero-vector"
**Test script:**
Problem is, when you try to normalize a Vector Object, which only has zero value, it tries to divide
them by 0 (their current magnitude), setting the values to ```nan```. It would be handy if a error would pop up.
```python
from p5 import *

x = Vector(0, 0)

print(x.magnitude)
x.normalize()
print(x)
print(x.magnitude)
```

**Expected behavior:**
```
0.0
ValueError(...)
```

**Actual behavior**
```
0.0
Vector(nan, nan, nan)
nan
```

**Errors produced (if any):**

**p5 version:**
0.3.0a2

**Python version:**
Python 3.6.4

**Operating System:** 
MacOS

parsoyaarihant:
Fixes #37.

Added test condition in line 376 to raise ValueError when magnitude of vector is 0.

```python 
>>> from p5 import *
>>> x = Vector(0, 0)
>>> print(x.magnitude)
0.0
>>> x.normalize()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/arihantparsoya/Documents/p5/p5/pmath/vector.py", line 368, in normalize
    raise ValueError('Cannot normalize vector with magnitude 0')
ValueError: Cannot normalize vector having magnitude 0
```

### DIFF
--- a/p5/pmath/vector.py
+++ b/p5/pmath/vector.py
@@ -364,6 +364,8 @@ class Vector(Point):
 
     def normalize(self):
         """Set the magnitude of the vector to one."""
+        if self.magnitude == 0.0:
+            raise ValueError('Cannot normalize vector with magnitude 0')
         self.magnitude = 1
 
     def limit(self, upper_limit=None, lower_limit=None):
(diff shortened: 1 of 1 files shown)


abhikpal:
Looks good :+1: Have merged into ```develop```

Issue #23, opened 2017-12-25
Tuple unpacking incompatible with Python 3.4 and less.


abhikpal:
```python
    from p5 import *
  File "/Users/arihantparsoya/Documents/p5/p5/__init__.py", line 19, in <module>
    from .sketch import *
  File "/Users/arihantparsoya/Documents/p5/p5/sketch/__init__.py", line 19, in <module>
    from .base import *
  File "/Users/arihantparsoya/Documents/p5/p5/sketch/base.py", line 29, in <module>
    from ..opengl import renderer
  File "/Users/arihantparsoya/Documents/p5/p5/opengl/__init__.py", line 19, in <module>
    from .renderer import *
  File "/Users/arihantparsoya/Documents/p5/p5/opengl/renderer.py", line 33, in <module>
    from ..pmath import matrix
  File "/Users/arihantparsoya/Documents/p5/p5/pmath/__init__.py", line 22, in <module>
    from .curves import *
  File "/Users/arihantparsoya/Documents/p5/p5/pmath/curves.py", line 46
    ret_value = func(*new_args, parameter, **kwargs)
                                       ^
SyntaxError: only named arguments may follow *expression
```

If tuple unpacking is something we really need we'll have to make Py3.5 a dependency, else we figure out a way to make this backwards compatible.

See #18 and #22 for more context. Also: thanks @parsoyaarihant for pointing this out!

PR #22, opened 2017-12-24
Replaced @ with np.dot()

parsoyaarihant:
Fixes #18 .

Replaced all `@` operator with `np.dot` since `@` operator was introduced from python3.5. 
There are still some issues while running the code:

```python 
    from p5 import *
  File "/Users/arihantparsoya/Documents/p5/p5/__init__.py", line 19, in <module>
    from .sketch import *
  File "/Users/arihantparsoya/Documents/p5/p5/sketch/__init__.py", line 19, in <module>
    from .base import *
  File "/Users/arihantparsoya/Documents/p5/p5/sketch/base.py", line 29, in <module>
    from ..opengl import renderer
  File "/Users/arihantparsoya/Documents/p5/p5/opengl/__init__.py", line 19, in <module>
    from .renderer import *
  File "/Users/arihantparsoya/Documents/p5/p5/opengl/renderer.py", line 33, in <module>
    from ..pmath import matrix
  File "/Users/arihantparsoya/Documents/p5/p5/pmath/__init__.py", line 22, in <module>
    from .curves import *
  File "/Users/arihantparsoya/Documents/p5/p5/pmath/curves.py", line 46
    ret_value = func(*new_args, parameter, **kwargs)
                                       ^
SyntaxError: only named arguments may follow *expression
```

PR #40, opened 2018-03-22
fix bug #23: Tuple unpacking incompatible with Python 3.4 and less

hx2A:
Simple fix for this bug so the package is backwards compatible.

### DIFF
--- a/p5/pmath/curves.py
+++ b/p5/pmath/curves.py
@@ -41,9 +41,9 @@ def typecast_arguments_as_points(func):
     """Typecast all but the last argument of the function as Points."""
     @wraps(func)
     def decorated(*args, **kwargs):
-        parameter = args[-1]
         new_args = [Point(*arg) for arg in args[:-1]]
-        ret_value = func(*new_args, parameter, **kwargs)
+        new_args.append(args[-1])
+        ret_value = func(*new_args, **kwargs)
         return ret_value
     return decorated
 


abhikpal:
Yep, this doe s fix the issue :) I've merged this to ```develop```

jeremydouglass:
More compatibility in the code base is better... but isn't it the case that p5py will still require minimum 3.5 https://github.com/p5py/p5/issues/41  anyway? At least, as long as it uses things like the [@ matrix multiplication operator](https://www.python.org/dev/peps/pep-0465/) -- as for example in:

https://github.com/p5py/p5/blob/master/p5/pmath/vector.py#L345

hx2A:
@jeremydouglass that is a fair point but I'd argue that the new version is also more readable. Also, if supporting previous versions of Python is important it would be easy to change the few places that use the `@` operator with the numpy matrix multiplication function.

jeremydouglass:
@hx2A -- good points. To be clear, I'm not against this change in #40 -- I agree that more previous Python 3.x version compatibility is better.

hx2A:
@jeremydouglass agreed! I like the new `@` operator also but perhaps if its use in p5py limits Python compatibility then that should be modified also? Replacing it with numpy's matrix multiplication function should suffice.

This project is part of the Google Summer of Code. Are one of you the person working on this?

abhikpal:
> This project is part of the Google Summer of Code. Are one of you the person working on this?

I'm working on this actually. PR #22 already solves the ```np.dot``` / ```@``` issue and it has been merged to the ```develop``` branch. These will be merged to ```master``` soon.

hx2A:
@abhikpal awesome! I'm eager to work with the final result at the end of the summer.

abhikpal:
@hx2A, thanks! You're welcome to test things out and give us feedback over the summer as well :)

PR #43, opened 2018-04-12
Added stroke_weight()

GOMMB:
There is probably a much better way of doing this, but this works for me so I just wanted to post it. Currently it only affects lines.


GOMMB:
I am trying to implement it a different way using `glLineWidth()` but I just get lost in your code.

PR #46, opened 2018-04-24
Implementation of DateTime and PShape

AliabbasMerchant:
This code was written by me as a prototype for my GSoC 2018 proposal.
I was not selected, but I would want you to please merge this with the main repo.
The code contains:
1. A complete DateTime module, similar to the Processing Java Code.
2. A basic prototype for the PShape module.
3. A demo of Pshape (PShape_demo.py)


PShape_Demo.py
```diff
@@ -0,0 +1,29 @@
+from p5 import *
```
abhikpal: We already have a repository for examples! This should ideally go there: https://github.com/p5py/p5-examples

p5/core/PShape.py
```diff
@@ -0,0 +1,157 @@
...
+POLYGON = 20  # in the end, probably cannot
+PATH = 21  # separate these two
+RECT = 30  # primitive
+ELLIPSE = 31  # primitive
+ARC = 32  # primitive
+SPHERE = 40  # primitive
+BOX = 41  # primitive
+
+
+class PShape():
```
abhikpal: We're dropping the ```P``` prefix in p5. So Processing's "PShape" becomes "Shape" in p5. We already have a ```Shape``` class and this new code should go in there.

p5/core/PShape.py
```diff
@@ -0,0 +1,157 @@
...
+
+from . import primitives
+from ..pmath import Point, Vector
+import warnings
+
+MIN_POINT_ACCURACY = 20
+MAX_POINT_ACCURACY = 200
+POINT_ACCURACY_FACTOR = 10
+
+OPEN, CLOSE = 1, 2
```
abhikpal: Is there a nicer way of doing this other than assigning numbers to constants?

p5/core/PShape.py
```diff
@@ -0,0 +1,157 @@
...
+        self._filled = True
+        self._fill = (0, 0, 0)
+        self._visible = True
+        self._openShape = False
+        self._kind = type
+        self._p = p
+        self._vertices = []
+        self.close = False
+
+    def beginShape(self, kind = None):
```
abhikpal: We're trying to use context managers in p5. So instead of 
```
foo.beginShape()
# code
foo.endShape()
```
we would like to have
```
with foo.begin_shape():
    # code
````
See the code for ```push_matrix()``` in p5/transforms.py for the implementation details. And the contextlib docs: https://docs.python.org/3/library/contextlib.html

File: p5/core/PShape.py [hunk 5]
```diff
@@ -0,0 +1,157 @@
...
+        self._vertices = []
+        if type is not None:
+            self._kind = type
+        self._openShape = True
+
+    def endShape(self, mode=OPEN):
+        if self._kind == GROUP:
+            warnings.warn("Cannot end GROUP shape")
```
abhikpal: Throwing an error here makes more sense, no?

p5/core/PShape.py
AliabbasMerchant: Yes.
That would be more pythonic.
My main aim was initially just to make it function just like the Java code.

abhikpal:
This is quite nice! But a couple of general comments:
- We have an example repository already, so the PShape_Demo should go there ideally: https://github.com/p5py/p5-examples
- I should probably put this some where in the contributor guidelines, but would be nice have PEP8 friendly code (```begin_shape()``` instead of ```beginShape()```) :) See: https://www.python.org/dev/peps/pep-0008/

p5/core/PShape.py
abhikpal: >  My main aim was initially just to make it function just like the Java code.

We don't have to. We're just keeping the higher-level api similar, the internals can definitely change! And of course, we always try to use some of the Python features whenever possible :)

p5/core/PShape.py
AliabbasMerchant: We may use strings instead.  
But numbers are faster to work with.  
Also, the variables have been named quite appropriately.  
Plus, these are the same constant names as used in the Java code, so there is a less of a learning curve/ less work while switching from Java to p5.

p5/core/PShape.py
AliabbasMerchant: Ok.  I'll keep that in mind.  
This is the same reason why I used `beginShape()` instead of `begin_shape()`

abhikpal: :+1:

p5/core/PShape.py
abhikpal: I was wondering for other solutions that do not necessarily use these variables. For instance, instead of defining ```OPEN``` and ```CLOSE``` we could just use keyword arguments: ```def end_shape(self, open=True)```

> We may use strings instead.  But numbers are faster to work with.

Correct. But this doesn't seem to influence things that much. I looked around a bit and at least in the popular scientific computing libraries this seems to be quite common. In scipy for instance, one would often do ```scipy.optimize.minimize(f, x0, method='CG')```. I really don't think something like ```self._kind == 'group'``` would be an issue.

p5/core/PShape.py
```diff
@@ -0,0 +1,157 @@
...
+        self._visible = True
+        self._openShape = False
+        self._kind = type
+        self._p = p
+        self._vertices = []
+        self.close = False
+
+    def beginShape(self, kind = None):
+        self._vertices = []
+        if type is not None:
```
abhikpal: ```type``` is a Python built-in!

p5/core/PShape.py
GoToLoop: Seems like if n1 attempts to convert a Java or Python Mode sketch into p5py is gonna know hell!  :P

abhikpal: @GoToLoop, ¯\\\_(ツ)\_/¯

p5/core/PShape.py
AliabbasMerchant: Sorry. My bad😌

auraz:
Please resolve conflicts with main branch

parsoyaarihant:
The features in this pull request have been incorporated in the current version of p5py

PR #49, opened 2018-05-24
Fixes retina issues on the mac

antiboredom:
This also includes the framerate fix by @parsoyaarihant.

### DIFF
--- a/p5/sketch/base.py
+++ b/p5/sketch/base.py
@@ -41,16 +41,10 @@ last_recorded_time = time.time()
 builtins.pixel_width = 1
 builtins.pixel_height = 1
 
-platform = pyglet.window.get_platform()
-display = platform.get_default_display()
+display = pyglet.canvas.get_display()
 screen = display.get_default_screen()
-
-template = pyglet.gl.Config(samples_buffers=1, samples=2)
-try:
-    config = screen.get_best_config(template)
-except pyglet.window.NoSuchConfigException:
-    template = pyglet.gl.Config()
-    config = screen.get_best_config(template)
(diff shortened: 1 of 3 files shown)


Manindra29:
I tried it on my machine today. I ran a bunch of sketches in the examples. The frame rates look much better! This is great @antiboredom!

But I did notice some screen tearing. For example in [interpolate.py](https://github.com/p5py/p5-examples/blob/master/basics/math/interpolate.py). Did anyone else notice the same? Not sure if this PR necessarily causes it, because frame rates were too low on Mac for the tearing to be noticeable without this PR.

abhikpal:
> But I did notice some screen tearing. For example in interpolate.py. Did anyone else notice the same?

I don't notice any tearing. That can be worked on later, though. At least the sketches now work on Mac :)

p5/sketch/base.py
```diff
@@ -62,6 +56,10 @@
     config=config,
 )
 
+actual_width, actual_height = window.get_viewport_size()
```
abhikpal: This seems to break the documentation auto-generation (see #50) I've pushed a fix on the ```fix-50-doc-autogen``` branch, can you see if it still works as expected on the Mac?

abhikpal:
@Manindra29, @antiboredom, could you try testing out the fix in the documentation generation (see #50) in ```fix-50-doc-autogen``` on your respective computers?

Manindra29:
The documentation build passed with #50 on my machine.

antiboredom:
@abhikpal I wasn't able to build the docs, but I think it's related to me messing up my python installation. That said everything else seems to work...

PR #58, opened 2018-08-01
Fix typo on HSB mode with alpha

jrembold:
Stumbled across this issue recently using HSB color mode with an alpha value. Figured it was a rare enough of a situation that no one had caught the typo yet and it was an easy fix.


abhikpal:
Ahh.. a tiny typo did creep in there. Thanks for finding and fixing this. I've merged it into `develop` and this will be part of the next release. 

Thanks!

PR #61, opened 2018-08-07
Small corrections/improvements

mark-summerfield:
Note that this file doesn't seem to be linked from any other in the docs: I only found it browsing the repo.


abhikpal:
Thanks! I've merged this to `develop` : )

PR #65, opened 2018-08-08
[FIX] failed exit() due to uninitialized default_sketch

### LINKED ISSUE #64: Error due to calling show() on uninitialized default_sketch.
**Test script:**

```python
# put your test script here
from p5 import *
exit()
```

**Expected behavior:**
Should actually exit.

**Actual behavior**
Does not exit()

**Errors produced (if any):**
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jessop/git/vendor/p5/p5/sketch/userspace.py", line 213,
in exit
    default_sketch.show(visible=False)
AttributeError: 'NoneType' object has no attribute 'show'
```
**p5 version:**
'0.4.0a1.dev2' from git 6d82eb85a160d1c5d36b6ef443b7b9e9d3f1e2cf

**Python version:**
3.6.5
**Operating System:** 
Manjaro

jessopb:
Fixes #64

### DIFF
--- a/p5/sketch/userspace.py
+++ b/p5/sketch/userspace.py
@@ -210,7 +210,11 @@ def exit(*args, **kwargs):
     :param kwargs: keyword-arguments to pass to Python's builtin
         `exit()` function.
     """
-    default_sketch.show(visible=False)
+    try:
+        default_sketch.show(visible=False)
+    except AttributeError:
+        pass
+
     app.quit()
     builtins.exit(*args, **kwargs)
 


p5/sketch/userspace.py
```diff
@@ -210,7 +210,11 @@ def exit(*args, **kwargs):
     :param kwargs: keyword-arguments to pass to Python's builtin
         `exit()` function.
     """
-    default_sketch.show(visible=False)
+    try:
```
abhikpal: I would avoid a try catch.. just in the odd chance that something else throws an `AttributeError`. A simple check for `default_sketch == None` should suffice.

abhikpal:
Thanks for finding the bug and doing the pull request. As I've included a slightly different fix (that doesn't raise an `AttributeError`) on the `develop` branch. I'm closing this now.

PR #74, opened 2018-11-15
Fix typo in Processing user guide: event.key() to event.key

1deadpixl:
[Event System](https://p5.readthedocs.io/en/latest/guides/for-processing-users.html#event-system) example was calling event.key() instead of referencing the attribute event.key.


abhikpal:
Thanks! Have merged to `develop`

Issue #86, opened 2019-01-12
Allow applying normalize() method right after Vectors +-*/ operations


jeremy886:
**Allow applying normalize() method right after Vectors +-*/ operations**

Based on the example given below, you cannot apply normalize() right after doing - operation. It's very counter intuitive for Python.

```
self.acc = (mouse - self.pos).normalize()  # this won't work
```

instead, I need to do this:

```
self.acc = mouse - self.pos
self.acc.normalize()  # this is okay
```

**Describe the solution you'd like**
Enable normalize() just like ("abc" + "def").upper()

**Describe alternatives you've considered**
Keep the same (similar to p5js) but mention this in the documentation.

**Additional context**

Given this example:

```
class Walker:
    def __init__(self):
        self.pos = Vector(width/2, height/2)
        self.vel = Vector(0, 0)
        self.acc = Vector(0, 0)

    def display(self):
        fill(255)
        ellipse(self.pos, 48, 48)

    def update(self):
        mouse = Vector(mouse_x, mouse_y)
        self.acc = (mouse - self.pos).normalize()  # this is not okay
        # self.acc = mouse - self.pos
        # self.acc.normalize()  # this is okay
        self.vel += self.acc
        self.pos += self.vel


def setup():
    global w
    size(640, 360)
    background(51)

    w = Walker()


def draw():
    w.update()
    w.display()


if __name__ == '__main__':
    run(frame_rate=30)
```

jeremy886:
Thanks. I created a PR

auraz:
Can we close this issue?

abhikpal:
@auraz, yep. This has been merged to `develop`

PR #87, opened 2019-01-13
allow chaining "normalize()" after +/- operations

jeremy886:
Like this: `self.acc = (mouse - self.pos).normalize()`

Addresses #86


auraz:
@abhikpal @jeremy886 please core PR once it is merged

Issue #118, opened 2019-08-14
triangle module missing  in development branch


SushilRagoonath:
i'm trying to import the latest developer branch of the code and the triangle module is missing in the core folder .

parsoyaarihant:
Hi @SushilRagoonath , can you elaborate on the issue? If you are getting an error, can you post it here?

SushilRagoonath:
File "<stdin>", line 1, in <module>
  File "D:\Users\Sushil\Downloads\p5-develop\p5\__init__.py", line 19, in <module>        
    from .sketch import *
  File "D:\Users\Sushil\Downloads\p5-develop\p5\sketch\__init__.py", line 21, in <module> 
    from .userspace import *
  File "D:\Users\Sushil\Downloads\p5-develop\p5\sketch\userspace.py", line 31, in <module>
    from .base import Sketch
  File "D:\Users\Sushil\Downloads\p5-develop\p5\sketch\base.py", line 32, in <module>     
    from ..core import p5
  File "D:\Users\Sushil\Downloads\p5-develop\p5\core\__init__.py", line 18, in <module> 
    from .structure import *
  File "D:\Users\Sushil\Downloads\p5-develop\p5\core\structure.py", line 21, in <module>
    from . import primitives
  File "D:\Users\Sushil\Downloads\p5-develop\p5\core\primitives.py", line 34, in <module>
    from .shape import PShape
  File "D:\Users\Sushil\Downloads\p5-develop\p5\core\shape.py", line 26, in <module>
    import triangle as tr
ModuleNotFoundError: No module named 'triangle'

SushilRagoonath:
Thats the full output I got,thanks for the help.

parsoyaarihant:
Right now, you need to install the triangle library separately - `pip install triangle`

SushilRagoonath:
Okay,any ETA on it being implemented in this project?

abhikpal:
@SushilRagoonath, probably when we prepare the release / merge to master. Only `setup.py` and `requirements.txt` will have to be updated.

SushilRagoonath:
All working now ,thanks.

PR #119, opened 2019-08-16
add triangle to requirements.txt, setup.py

jeremydouglass:
add triangle to install dependencies as per #118


parsoyaarihant:
Thank you @jeremydouglass !

PR #127, opened 2019-09-15
fix more tutorial examples

edurojasr:
fixed the "coordinate system and shapes" example at the end of the tutorial so now works as intended 
fixed the "Mouse Data" print x and y mouse position
fixed some text format problems in "Interactivity" tutorial
fixed the angle example of interactivity tutorial, added ord and str methods
in Coded keys example is not necessary define "y" value as 35 outside def 
fixed the "mouse_moved()" and "mouse_dragged()" example in the Interactivity Tutorial


edurojasr:
I'm going to close this pull request because I thought I'm going to make a better one containing more changes and better commented

PR #128, opened 2019-09-16
More tutorials examples fixed

edurojasr:
**From Coordinate System and Shapes Tutorial:**
- fixed the "coordinate system and shapes" example at the end of the tutorial so now works as intended

**From Interactivity Tutorial:** 
- fixed the "Mouse Data" examples print x and y mouse position
- fixed some text format problems in "Interactivity" tutorial
- in "Keyboard data" fixed the angle example of interactivity tutorial, added ord() and str() methods
- in "Coded keys" example is not necessary define "y" value as 35 outside def
- in "Mouse events" example fixed the "mouse_moved()" and "mouse_dragged()" example

**From Typography Tutorial:**
- in "Vector Fonts" example fixed text format and change "textFont()" to "text_font()"
- in "Vector Fonts" in ...To use two fonts added global and deleted unnecessary text_font
- in "Vector Fonts" example deleted unnecessary variables
- in "Typing" second example improved so its works better

**Edit**

Adding @jeremydouglass suggestions  (Thanks! btw)
 1.  Making variables more easy to understand in the interactivity tutorial
 2. Declaring global variables in the typography tutorial
 3. Deleting commented out "text_size()"

### DIFF
--- a/docs/tutorials/coordinate system and shapes.rst
+++ b/docs/tutorials/coordinate system and shapes.rst
@@ -105,11 +105,20 @@ Now let's look at what some code with shapes in more realistic setting, with win
 
 .. code:: python
 
-	size(200,200)
-	rectMode("CENTER")
-	rect((100,100),20,100)
-	ellipse((100,70),60,60)
-	ellipse((81,70),16,32) 
-	ellipse((119,70),16,32) 
-	line((90,150),(80,160))
-	line((110,150),(120,160))
+	from p5 import *
+
+
+	def setup():
+    	size(200, 200)
+
+
+	def draw():
+    	rect_mode(CENTER)
+		rect((100, 100), 20, 100)
+		ellipse((100, 70), 60, 60)
+		ellipse((81, 70), 16, 32)
+		ellipse((119, 70), 16, 32)
+		line((90, 150), (80, 160))
+		line((110, 150), (120, 160))
+	run()
+
--- a/docs/tutorials/coordinate system and shapes.rst
+++ b/docs/tutorials/coordinate system and shapes.rst
@@ -107,13 +107,11 @@ Now let's look at what some code with shapes in more realistic setting, with win
 
 	from p5 import *
 
-
 	def setup():
-    	size(200, 200)
-
+		size(200, 200)
 
 	def draw():
-    	rect_mode(CENTER)
+		rect_mode(CENTER)
 		rect((100, 100), 20, 100)
 		ellipse((100, 70), 60, 60)
 		ellipse((81, 70), 16, 32)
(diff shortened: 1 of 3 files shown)


docs/tutorials/typography.rst
```diff
@@ -175,8 +175,6 @@ Before a font is used in a program, it must be converted and set as the current
 
    from p5 import *
 
-   zigBlack = None
```
jeremydouglass: I think the first two examples should also declare zigBlack / sourceLight "global" in setup, even if it isn't strictly necessary for those example -- that is a good basic expectation, that these should be accessible later if you want them. Learners cut-paste these as boilerplate... and then later they try to use `textFont()` and it doesn't work and they don't understand why.

docs/tutorials/typography.rst
```diff
@@ -388,25 +388,32 @@ Drawing letters to the screen becomes more engaging when used in combination wit
...
    letters = ""
    back = 102
+   fil = 0
+
 
    def setup():
       size(100, 100)
       text_align("CENTER")
-      text_size(16)
+      # text_size(16)
```
jeremydouglass: I'd suggest either use it or don't, don't comment it out. If you don't like the size but want to use it, make it 10 or 12.

docs/tutorials/coordinate system and shapes.rst
```diff
@@ -105,11 +105,20 @@ Now let's look at what some code with shapes in more realistic setting, with win
...
+	from p5 import *
+
+
+	def setup():
+    	size(200, 200)
+
+
+	def draw():
+    	rect_mode(CENTER)
+		rect((100, 100), 20, 100)
```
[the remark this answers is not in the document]
edurojasr: yes! sorry about that

docs/tutorials/interactivity.rst
```diff
@@ -717,7 +717,7 @@ The code inside the ``mouse_moved()`` and ``mouse_dragged()`` event functions ar
 
    from p5 import *
 
-   dragX, dragY, moveX, moveY = (0, 0, 0, 0)
+   dragX = dragY = moveX = moveY = 0
```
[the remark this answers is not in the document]
edurojasr: Totally  agree, for new learners as you say its better to see each variable name with its value

docs/tutorials/typography.rst
edurojasr: Ok agreed, notice that textFont() is not correct ether, should us text_font() but I understand why its good idea to have declare zigBlack / sourceLight "global", maybe for me is weird having variables that are not used.

docs/tutorials/typography.rst
edurojasr: This is because I saw another example with "text_size()" comment it out, the example is in "Typography" tutorial, the last example have that function comment it out and I think it is related to the fact that the current default font is bitmap, so the text_size function is not yet implemented for that format.

edurojasr: I'm going to delete the "text_size" since is not implemented yet for bitmap fonts and because using it require make the font a vector and the example became a little to complex

PR #145, opened 2020-02-25
Correct the font file extension

tabreturn:


jeremydouglass:
This seems like a seem, obvious change that fixes a problem.
Is there a reason that it cannot be merged?

PR #147, opened 2020-03-27
[DOC] Remove extra pop_matrix() call in arm example

ziyaointl:
According to the documentation [here](https://p5.readthedocs.io/en/latest/guides/for-processing-users.html), `push_matrix` is a context manager and `pop_matrix` does not exist. In reality, `pop_matrix` exists but seems to do nothing according to the [source](https://github.com/p5py/p5/blob/1b233fdd59f19e97341e5cfb75bccb821ff79eab/p5/core/transforms.py#L41).

While a future PR is probably needed to clean up the implementation of the API, this PR removes the extraneous pop_matrix() call in the arm example code.


parsoyaarihant:
@ziyaointl Thanks!

PR #157, opened 2020-04-24
Fix push_matrix, drop pop_matrix and pop_style

### LINKED ISSUE #156: push_matrix() and pop_matrix() not working
using pop_matrix() does not restore the transform matrix to the previous state. I tried using the exact codes mentioned in your documentation : https://p5.readthedocs.io/_/downloads/en/latest/pdf/. 

from p5 import *
def setup():
    size(200, 200)
def draw():
    background(255)
    fill(192)
    no_stroke()
    rect((40, 40), 40, 40)
    push_matrix()
    translate(40, 40)
    rotate(radians(45))
    fill(0)
    rect((40, 40), 40, 40)
    pop_matrix()
if __name__ == '__main__':
    run()

The following code clearly depicts the error

jeremydouglass:
This removes pop_matrix() and pop_style(), which were unnecessary (push_matrix and push_style are context managers) _and_ were incorrectly implemented.

It also corrects all the bad documentation and example code, which was incorrectly demonstrating push_matrix as a non-context manager (wrong) followed by pop_matrix (wrong, shouldn't be used). This includes narrative explanations from the 2D tutorial.

Closes #156 
Resolves #147


jeremydouglass:
@tyagi619 wants to contribute a pull request on for #156, so I'll withdraw this one.

jeremydouglass:
@parsoyaarihant @abhikpal -- I'm requesting a review.

This should be very quick to look over and either approve or ask for changes.

PR #160, opened 2020-04-29
Closes #158 draw() being called multiple times and frame_count was incremented in wrong manner

### LINKED ISSUE #158: no_loop incorrectly allows an extra draw frame
**Describe the bug**
`no_loop` should stop draw from being called after the function in which it is called returns. Instead, it draws another frame, *then* stops draw.

**To Reproduce**
```
from p5 import *
def setup():
    size(400,400)
def draw():
    print("frame",frame_count)
    no_loop()
if __name__ == '__main__':
    run()
```

**Expected behavior**

Should print "frame 1".
Instead prints "frame 1 frame 2".

**System information:**
 - p5 release 0.6.0
 - Python version:3.7.7
 - Operating system: macOS 10.12

tushar5526:
In the base.py frame_count was incremented without checking whether or not looping is set to true or false
Draw() was called again because there was no check to see if looping is false and redraw is true, then don't call draw()
Fixed it by putting a elif not self.looping : pass in the if else statements, 

You can test to see if everything and frame_count are working fine by putting print(frame_count) in base.py at various positions in the code

### DIFF
--- a/p5/sketch/base.py
+++ b/p5/sketch/base.py
@@ -91,8 +91,8 @@ class Sketch(app.Canvas):
         builtins.frame_rate = round(self.fps, 2)
 
         with p5.renderer.draw_loop():
-            builtins.frame_count += 1
             if not self.setup_done:
+                builtins.frame_count += 1
                 self.setup_method()
                 self.setup_done = True
                 self.show(visible=True)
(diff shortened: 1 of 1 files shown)


parsoyaarihant:
@tushar5526 Thanks!

PR #164, opened 2020-05-08
+ Implement text stroke feature and speed up large image rendering

Hansimov:
Fixes:
* Speed up image rendering, especially large images

Added Features:
* Support rendering outline of text with `stroke()`
* Support 'inner' outline when `stroke_weight` is negative
* Do not render text when `no_fill()` is used

Core ideas:
1. Use PIL filter to 'dilate'  (`ImageFilter.MaxFilter(stroke_weight)`) or 'erode' (`ImageFilter.MinFilter(stroke_weight)`) original created `text_image`
2. Use PIL `ImageChops.difference(canvas,canvas_dilate)` to get 'outline' image
3. Render the original text and outline image successively

Example code here (unnecessary details are hidden):

```python
def draw():
    background(255)

    text_align("center","center")
    text_size(100)

    en_str = "hello 你好"
    fill(255,0,0)

    gap = 100
    height = 20
    text(en_str,(50,height))

    stroke(0,0,255)
    stroke_weight(5)
    height += gap
    text(en_str,(50,height))

    stroke_weight(-5)
    height += gap
    text(en_str,(50,height))

    no_fill()
    height += gap
    stroke_weight(5)
    text(en_str,(50,height))

    height += gap
    stroke_weight(-5)
    text(en_str,(50,height))
```
text_stroke


parsoyaarihant:
@Hansimov have you completed working on "large image rendering"?

Hansimov:
> @Hansimov have you completed working on "large image rendering"?

Yes, I have.

jeremydouglass:
Text stroke is a great idea!

Given that stroke_weight is already part of every other API and does something different -- and has a different valid argument range -- should the text stoke and text stroke weight be set with something separate, like `text_stoke_weight`?

I'd also suggest making this text stoke method a new mode, using `text_mode`. 

https://processing.org/reference/textMode_.html

Hansimov:
@jeremydouglass I have considered this issue.
Since Processing and p5.js both directly use `stroke()` to draw the outline of texts, I think it might be better to keep this consistency.

Of course you can add a new mode or method to control the stroke of text seperately.
By the way, if you really want to seperate the stroke of texts from the stroke of shapes, I suggest to wrap it into the parameters of `text()`, like this `text(txt_str,(x,y), is_stroke=True, stroke_weight=10)`. Using this method you can wholly seperate this stroke feature of text.

However, since the `fill()` function controls the colors of both shapes and texts, I still recommend to let `stroke()` manage the stroke drawing of both shapes and texts, just like what Processing and p5.js do.

jeremydouglass:
> Since Processing and p5.js both directly use stroke() to draw the outline of texts, I think it might be better to keep this consistency.

Unless we are misunderstanding each other, I think you are mistaken. Processing does not use `stroke()` (or `strokeWeight()`) to draw the outline of texts. Here is a sketch with a 10-pixel thick red stroke defined.

```
textSize(32);
fill(0, 102, 153);
strokeWeight(10);
stroke(255,0,0);
text("word", 10, 60);
```

Screen Shot 2020-05-09 at 10 31 25 PM

As you can see, `stroke()` and `strokeWeight()` are not used to draw the text.

However, it is true that p5.js does this -- I had forgotten that Processing and p5.js were inconsistent in this way, and this makes more sense in that context.

Hansimov:
@jeremydouglass Oh, You're right. 2333333
I mixed the Processing and p5.js up on this small detail. Thanks for pointing it out. :P

Issue #130, opened 2019-09-26
Why is the syntax different from processing


hunar1997:
I searched the readme and the issues and couldn't find my question, i just found p5py and tried to transfer a python mode project within processing which uses jython, and it didn't work. The error message was confusing, but i found that the reason was that in processing `circle(x,y,r)` is correct but in p5py it must be `circle((x,y),r)`

but why? wouldn't it be better to make the syntax the same? to attract processing users and make it easy for them?

Thanks :D

jeremydouglass:
Personally I think that seems like a reasonable expectation.

`func(a,b,c)` is generally the way you do it in Processing Java, processing.py, p5.js, Processing.R, JRubyArt et cetera -- circle is a recent addition in Processing 3.x, but ellipse() or rect() work the same way in everything -- not rect((x,y),w,h) but rect(x,y,w,h). This also makes sense with the various "mode" parts of the API, like ellipseMode and rectMode, which can change what the arguments do.

No reason you couldn't support both that AND PVectors / points / tuples / whatever signatures, but flat argument lists is the default in every other API.

parsoyaarihant:
> The error message was confusing, but i found that the reason was that in processing circle(x,y,r) is correct but in p5py it must be circle((x,y),r)

This is something I found confusion too. I believe we need to have a bigger discussion about it. 
Adding @abhikpal to the conversation.

> This also makes sense with the various "mode" parts of the API, like ellipseMode and rectMode, which can change what the arguments do.

The names of the '"mode" functions were originally created to match the Pythons [PEP 8](https://www.python.org/dev/peps/pep-0008/) conventions. Hence, the convention has persisted. We contacted Ben Fry regarding this issue and he approved the current PEP 8 convention.

swirly:
Found this very disturbing for teaching. Have to explain tuples far too soon to use this

GoToLoop:
> No reason you couldn't support both that AND PVectors / points / tuples / whatever signatures,

That's exactly what I did very recently in a Processing forum thread:
https://Discourse.Processing.org/t/get-perpendicular-angle/14067/14

Functions **line()**, **circle()** & **text()** were reimplemented w/ a new signature, but w/o losing the original 1.

For the **circle()**'s new implementation below, besides keeping its **_PApplet_**'s original 3-parameter signature, it also accepts any object containing attributes _x_ & _y_.

[...]

PR #169, opened 2020-05-25
Improve p5py API compatibility with Processing

ziyaointl:
This PR addresses #130 and is a work-in-progress. Feel free to comment if you have any suggestions!

In short, this PR will enable p5py to support non-tuple arguments in functions that have counterparts in the standard Python Mode of Processing.

### Implementation
I plan to use if statements to check the number of parameters being passed in.
For example:
```python
def line(*args):
    if len(args) == 2:
        p1, p2 = args[0], args[1]
    elif len(args) == 4:
        p1, p2 = args[:2], args[2:]
    elif len(args) == 6:
        p1, p2 = args[:3], args[3:]
    else:
        raise ValueError("Unexpected number of arguments passed to line()")
        
    <Rest of existing code>
    ...
```
Another more elaborate example with error checking
```python
from collections import Iterable

def line(*args):
    if len(args) < 2:
        # Not enough arguments are passed 
        raise ValueError("Invalid number of args to line; expected more than 2")
    elif not isinstance(args[0], Iterable):
        if len(args) == 4: # 2D
            p1, p2 = args[:2], args[2:]
        elif len(args) == 6: # 3D
            p1, p2 = args[:3], args[3:]
        else:
            raise ValueError("Invalid nubmer of args to line; expected 4 or 6 arguments when first argument is not an iterable")
    else:
        assert len(args) == 2, "Invalid number of args to line; expected two iterables when first element is an iterable"
    
    <Rest of existing code>
    ...
```
I am leaning towards the simple version because
1. It is cleaner
2. The code will fail later anyway if incorrect parameters are passed in

A con to the simple approach is that the error messages can be more cryptic if we rely on later functions to check the parameters for us. Therefore, which version to use is open for discussion.

### Documentation Changes
Add a section under each function to document new API calls as well as old ones explicitly. I found the format in the Proccesing Language Docs to be helpful.

Example: https://processing.org/reference/line_.html
Rendered in p5py docs:

### Functions to be changed
``shape.py``

- [ ] p5.line()
- [ ] p5.ellipse()
- [ ] p5.circle()
- [ ] p5.arc()
- [ ] p5.triangle()
- [ ] p5.quad()
- [ ] p5.rect()
- [ ] p5.square()
- [ ] p5.bezier()
- [ ] p5.bezier_point()
- [ ] p5.bezier_tangent()
- [ ] p5.curve()
- [ ] p5.curve_point()
- [ ] p5.curve_tangent()

``transforms.py``
- [ ] p5.camera()

``image.py``
- [ ] p5.image()

``font.py``
- [ ] p5.text()

### DIFF
--- a/p5/core/primitives.py
+++ b/p5/core/primitives.py
@@ -188,11 +188,33 @@ def point(x, y, z=0):
 def line(p1, p2):
     """Returns a line.
 
+    :Syntax: * line(p1, p2)
+             * line(x1, y1, x2, y2)
+             * line(x1, y1, z1, x2, y2, z2)
+
     :param p1: Coordinates of the starting point of the line.
     :type p1: tuple
 
     :param p2: Coordinates of the end point of the line.
     :type p2: tuple
+
+    :param x1: x-coordinate of the first point
+    :type x1: float
+
+    :param y1: y-coordinate of the first point
+    :type y1: float
+
+    :param z1: z-coordinate of the first point
+    :type z1: float
+
+    :param x2: x-coordinate of the second point
+    :type x2: float
+
+    :param y2: y-coordinate of the second point
+    :type y2: float
+
+    :param z2: z-coordinate of the second point
+    :type z2: float
 
     :returns: A line PShape.
     :rtype: PShape


jeremydouglass:
Thanks, this is really great! Both work, but I also lean towards the first style.

My thinking: Helpful error messages are important for learners, but it isn't clear to me that the more detailed error messages are significantly more helpful in this case -- the user still may need to look up the expected argument signatures either way. Given the size of the API to address, starting with very a simple clean set of mapping seems like a better first step. If that all works well, it could be elaborated later in a more-helpful-messages phase.

For example, I believe that the demo of the second style already contains an incorrect error message (wrong: "expected more than 2" correct: "expected 2 or more" / "expected at least 2") -- and these kinds of things can be hard to catch on a big API surface without a lot of testers.

ziyaointl:
Hi Jeremy,

Thanks for your kind words. I agree with your reasons on why we should choose the simple approach as well as the plan to add more situation-aware error messages later, if needed.

ziyaointl:
Oops, I accidentally force-pushed to master while rebasing this branch. I quickly restored master via `git reflog`,  but this PR seems to be irreversibly closed.

Apologies for any confusion that this action has caused. To maintain one source of truth, I will still use this PR to track my progress (fortunately the checkboxes still work). When the code is ready to be reviewed, I will open another PR and link it back to this one.

A huge lesson learned not to ever force-push to repositories that have more than one developer. Other constructive criticism is extremely welcome. Again, sorry!

PR #176, opened 2020-05-31
Correct implementation of no_loop()

### LINKED ISSUE #175: no_loop() when used in setup() don't call draw() but it should call draw() only once
**Describe the bug**
According to processing https://processing.org/examples/noloop.html `The noLoop() function causes draw() to only execute once. Without calling noLoop(), the code inside draw() is run continually.`, 
In p5py no_loop() don't call the draw() function once

**To Reproduce**
```
from p5 import *

def setup():
    background(255)     # Set stroke color to white
    no_loop()

def draw():
    print('called draw')

if __name__ == '__main__':
    run()
```
**Current Behaviour**
No output

**Expected behavior**
It should print `called draw`

**System information:**
 - p5 : cloned from github

tushar5526:
Closes #175 

#158 had problem of draw() being called two times if we use no_loop() inside draw() function, which was fixed by a PR by me but it led to #175, that is no_loop() when used in setup() don't call draw() once, as per Processing original documentation

This PR calls draw() once when used in setup() and don't call draw() again if no_loop() is used in draw() again, also I have tested the frame_count for each call and frame_count is being incremented as described by @jeremydouglass  in #162 comments,  

The issue #158 and #162 is taken care by this PR , I have tested the code for loop(), no_loop(), redraw() for both in setup() and draw() functions and some other codes from examples of p5 

But you can test it for some other scenarios too :)

PS : size() when used in setup() along with no_loop() is still buggy, but I don't think it is happening due to no_loop()

### DIFF
--- a/p5/sketch/base.py
+++ b/p5/sketch/base.py
@@ -100,14 +100,17 @@ class Sketch(app.Canvas):
                     self.redraw = False
                 if self.looping is None:
                     self.looping = True
+
+            elif self.redraw:
+                builtins.frame_count += 1
+                self.draw_method()
+                self.redraw = False
             elif self.looping:
                 builtins.frame_count += 1
                 self.draw_method()
+                self.redraw = False
             elif not self.looping:
                 pass
-            elif self.redraw:
-                self.draw_method()
-                self.redraw = False
 
             while len(self.handler_queue) != 0:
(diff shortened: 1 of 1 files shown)


parsoyaarihant:
Thanks!

PR #184, opened 2020-06-18
Example does not work

ReneTC:
Old tutorial did not work. It also didn't match the image shown. I fixed the code to match the image. Also, I changed the name from "inside", "middle" and "outside" since these names are nonsense.


parsoyaarihant:
Thanks

ReneTC:
np @parsoyaarihant - I love this package but I'm a beginner to contributing. Please let me if there are any "easy" stuff I can do for this package.

PR #185, opened 2020-06-20
Add Pillow to requirements.txt

ziyaointl:
Pillow is used here: https://github.com/p5py/p5/blob/975f07e75a48aa7ebc8225a9cb5b27fda8795bb8/p5/sketch/base.py#L23

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ numpy
 glfw
 vispy
 triangle
+Pillow


jeremydouglass:
Should the requirements.txt be versioned from a `pip freeze`? Using un-versioned requirements seems like a recipe for surprise build-breaking....

ziyaointl:
> Should the requirements.txt be versioned from a `pip freeze`? Using un-versioned requirements seems like a recipe for surprise build-breaking....

Good point! I'll close this PR for now and open a new one with `pip freeze` soon.

PR #187, opened 2020-06-20
Update requirements.txt to include versions

ziyaointl:


parsoyaarihant:
What is the use of `freetype-py `?

ziyaointl:
> What is the use of `freetype-py `?

I believe that's a dependency of vispy
```bash
❯ pipdeptree                                                                                                                             p5 09:54:27 AM
glfw==1.11.2
Pillow==7.1.2
pipdeptree==1.0.0
  - pip [required: >=6.0.0, installed: 20.1.1]
PyOpenGL==3.1.5
PyOpenGL-accelerate==3.1.5
setuptools==47.3.1
triangle==20200424
  - numpy [required: Any, installed: 1.18.5]
vispy==0.6.4
  - freetype-py [required: Any, installed: 2.1.0.post1]
  - numpy [required: Any, installed: 1.18.5]
```

jeremydouglass:
On the one hand, requirements.txt is easier to read without sub-dependencies. We can only list direct imports, and sub-dependency versioning is handled by those projects -- so it is up to the vispy install how to handle freetype install.

On the other hand, if the dependency is portable then explicit naming it may be better. vispy doesn't actually specify a freetype-py version -- so there is no reason not to record the actual freetype-py version used that worked in development and testing.

It is also sometimes helpful to know about a major sub-dependency. numpy is a good example -- it might be interesting for people looking at requirements.txt and see that numpy will be required, even if it is not required directly.

ziyaointl:
@jeremydouglass Thanks for sharing some of the nuances in constructing a requirements.txt file. This is really helpful!

PR #194, opened 2020-07-09
Abstract common rendering logic in OpenGLRenderer

ziyaointl:
OpenGLRenderer is an abstract class that contains some of the common logic shared by the renderers that use vispy/opengl. It cannot be instantiated directly and is inherited by both Renderer2D and Renderer3D.


parsoyaarihant:
Good solution

PR #200, opened 2020-07-20
Implement `normal_shader`

ziyaointl:
Implements `normal_shader` that colors fragments based on the surface normal of a 3D model.

Sample:
```python
from p5 import *

def setup():
    size(720, 400)
    normal_material()

def draw():
    background(205, 102, 94)
    rotate_x(frame_count * 0.01)
    rotate_y(frame_count * 0.01)
    cone(200, 400)

if __name__ == '__main__':
    run(mode='P3D')
```

Screen Shot 2020-07-20 at 12 14 52 AM

### DIFF
--- a/p5/sketch/renderer3d.py
+++ b/p5/sketch/renderer3d.py
@@ -18,13 +18,14 @@
 
 import numpy as np
 from numpy.linalg import inv
+from enum import Enum
 import math
 from ..pmath import matrix
 
 import builtins
 
 from vispy import gloo
-from vispy.gloo import Texture2D
+from vispy.gloo import Texture2D, Program
 
 from contextlib import contextmanager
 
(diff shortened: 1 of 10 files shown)


p5/sketch/renderer3d.py
```diff
@@ -33,13 +34,22 @@
...
-from .openglrenderer import OpenGLRenderer, get_render_primitives
-from .shaders3d import src_default, src_fbuffer
+from .openglrenderer import OpenGLRenderer, get_render_primitives, to_3x3
+from .shaders3d import src_default, src_fbuffer, src_normal
+
+
+class Shader(Enum):
+	BASIC = 'BASIC'
+	NORMAL = 'NORMAL'
+	BLINN_PHONG = 'BLINN_PHONG'
```
parsoyaarihant, line 44: Should this be in separate class for reflection model?

p5/sketch/renderer3d.py
```diff
@@ -141,11 +159,12 @@ def render(self, shape):
 				np.hstack([shape.vertices, np.ones((n, 1))]),
 				shape.matrix,
 				self.transform_matrix)
+			tnormals = shape.vertex_normals @ np.linalg.inv(to_3x3(self.transform_matrix) @ to_3x3(shape.matrix))
```
parsoyaarihant, line 162: Would having a different `render` function for different materials be more efficient?

ziyaointl: That's a good point. I'll think about how to implement it.

p5/sketch/renderer3d.py
ziyaointl: I think (glossy) reflection can be covered in one blinn_phong shader. Could you elaborate on what may be separated?

parsoyaarihant: I was thinking of different classes for materials and reflection models. In the future, if they have same structure, then its fine to combine them in the same `Shader` class.

PR #203, opened 2020-07-25
Implement basic_material

ziyaointl:
Implements `basic_material`. More importantly, shader type is no longer a global state within `renderer3d`. This means we can have objects with different materials being rendered in the same frame.

Example:
```python
from p5 import *

def setup():
    size(720, 400)

def draw():
    background(205, 102, 94)
    normal_material()
    cone(200, 400)
    translate(200, 0, 0)
    rotate_x(radians(180))
    basic_material(155, 184, 48)
    cone(200, 400)

if __name__ == '__main__':
    run(mode='P3D')
```
Screen Shot 2020-07-25 at 4 13 14 AM

### DIFF
--- a/p5/core/geometry.py
+++ b/p5/core/geometry.py
@@ -20,6 +20,7 @@
 3D geometry class for p5py
 """
 
+from . import p5
 import numpy as np
 import math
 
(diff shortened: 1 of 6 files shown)


ziyaointl:
Sneak peak for the Blinn-Phong shader \:D

I'll open a PR as soon as the two pending ones are merged (due to code dependencies)
Screen Shot 2020-07-28 at 1 20 52 AM

Issue #183, opened 2020-06-18
Scalling with scale(1,0) does nothing. but scale(0,1) works  


ReneTC:
**Describe the bug**
recently I wanted to squish a box flat. Let's try it.
```

from p5 import *

def setup():
        size(640, 360)

def draw():
    no_stroke()

    background(0)

    fill(255)
    #here i will put different scale values
    rect((0, 0), 200, 200)


if __name__ == '__main__':
  run()

````
I get a white box:
![image](https://user-images.githubusercontent.com/40211569/85075570-c11c9580-b1be-11ea-9c79-400287f87c8c.png)

scaling with scale(0,1) flattens the box 100% in the x direction. Which makes perfect sense. But scaling with scale(1,0) does nothing. Box remains untouched. 

Everything works perfectly with other values for y, say scale(1,0.5). It squishes the box 50%. 

**Expected behavior**
y values should be smoshed to 0.

ziyaointl:
Resolved by #224

PR #224, opened 2020-08-07
Fixed issue #183

galaxyproduction:
The intended behavior was to scale sy and sz the same about as sx, if sx was the only variable not None. `not 0` returns True in python. That is why `scale(0, 1)` worked but `scale(1, 0)` didn't.


ziyaointl:
Thanks for contributing!

PR #227, opened 2020-08-09
Update setup.py

### LINKED ISSUE #222: Update installation scripts
After the removal of `triangle` and the introduction of `PyOpenGL`, the `setup.py` file is now out-of-date. Fix that.

Andy-Python-Programmer:
Fixes #222

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,15 +10,15 @@ here = os.path.abspath(os.path.dirname(__file__))
 
 packages = ['p5']
 
-requires = [
-    'vispy',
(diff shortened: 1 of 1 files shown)


ziyaointl:
Thanks a lot! The interplay between requirements.txt and setup.py seem to be a bit more [nuanced](https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py), but this should work for now.

PR #234, opened 2020-08-12
Implement scale_colors

ziyaointl:
The Color class imposes an upper bound to colors, which constrains
the maximum intensity of the lights. To offer users more freedom, a
separate scale_colors function was implemented that allows color channels
to go beyond 255.

### DIFF
--- a/p5/sketch/util.py
+++ b/p5/sketch/util.py
@@ -9,3 +9,8 @@ def read_shader(filename):
 
 def ensure_p3d(name):
     assert p5.mode == 'P3D', name + "is only available in P3D renderer"
+
+def scale_colors(r, g, b):
+    """Divides each channel by 255
+    """
+    return (r / 255.0, g / 255.0, b / 255.0)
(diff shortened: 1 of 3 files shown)


p5/sketch/util.py
```diff
@@ -9,3 +9,8 @@ def read_shader(filename):
 
 def ensure_p3d(name):
     assert p5.mode == 'P3D', name + "is only available in P3D renderer"
+
+def scale_colors(r, g, b):
```
parsoyaarihant, line 13: Can we use another name for this? Its not clear what scale is going to be applied. And it ignores the alpha channel.

ziyaointl: would `div_by_255` work? I guess we can also call it `unbounded_normalize`.

parsoyaarihant: We can use `div_by_255`

parsoyaarihant: Or add a default attribute `scale_tuple(tuple, scale=255):`

PR #236, opened 2020-08-12
Add example on Blinn Phong material

ziyaointl:
download


ziyaointl:
Had to do some hacks in the JavaScript code to make the looks consistent.

Issue #244, opened 2020-08-26
IO module in p5py


nakul-shahdadpuri:
**Is your feature request related to a problem? Please describe.**
Yes,
**P5py** currently doesn't have an IO module. I would like to work with live data/APIs/serial-input from my python processing sketch. 

**Describe the solution you'd like**
I would like functions in sync with [IO module of p5.js](https://p5js.org/reference/#group-IO) so that new users coming to p5py from p5.js do not face any difficulty.

**Describe alternatives you've considered**
Requests library can be used to perform GET/POST requests.

**Additional context**
I understand that most of these issues could be resolved through the requests library, thus if possible the IO module could also contain serial input (via USB) etc.

I would be interested to work on this project.

Andy-Python-Programmer:
Hi @nakul-shahdadpuri the IO link seems to a 404

nakul-shahdadpuri:
Hi @Andy-Python-Programmer 
For my first pull request I am thinking something around the lines of
[**httpPost()**](https://www.geeksforgeeks.org/p5-js-httppost-function/)
**[httpGet()](https://www.geeksforgeeks.org/p5-js-httpget-function/l)**
I ll try to keep the syntax similar to the javascript version. [(p5.js)](https://github.com/processing/p5.js)
Please tell me in which directory I can add these features and should I create a new **io** dir in p5/
Thank you.

Andy-Python-Programmer:
All of the p5py source is in the p5 folder @nakul-shahdadpuri . You might want to add IO directory but it depends if the folder only contains one file.

Andy-Python-Programmer:
Also while contributing follow the pep8 style:
httpGet should be http_get according to pep8

ziyaointl:
This is super exciting! Thanks for working on it.

Andy-Python-Programmer:
We could possibly also add read files and write files function also as for beginners and developers which are shifting from js it can be tough for them to remover w, r, r+ arg types.

nakul-shahdadpuri:
@Andy-Python-Programmer 
@ziyaointl 
P5.js has 
[loadJSON()](https://p5js.org/reference/#/p5/loadJSON)
[loadStrings()](https://p5js.org/reference/#/p5/loadStrings)
[loadTable()](https://p5js.org/reference/#/p5/loadTable)
[loadXML()](https://p5js.org/reference/#/p5/loadXML)
[loadBytes() ](https://p5js.org/reference/#/p5/loadBytes)
[...]

PR #245, opened 2020-08-28
Fixes Issue #244

nakul-shahdadpuri:
Regarding [Issue](https://github.com/p5py/p5/issues/244)
Added  Synchronous versions of **http_get()** and **http_post()**
Please let me know about any changes, and where should I update the documentation and examples.
Thank you. :)

### DIFF
--- a/p5/io/http.py
+++ b/p5/io/http.py
@@ -1,0 +1,2 @@
+def add(a,b):
+	return a+b
--- a/p5/io/http.py
+++ b/p5/io/http.py
@@ -1,9 +1,8 @@
-from urllib import request
+from urllib import request,parse
 import json
+import requests
 
-
-
-
+#synchronous
 def http_get(url):
 	try:
 		resp = request.urlopen(url)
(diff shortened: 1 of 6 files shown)


ziyaointl:
The changes look great!

For the documentation, we use sphinx and the readthedocs theme. To get started, install some denpendencies:
```bash
pip install sphinx sphinx_rtd_theme
```
Then you should be able to run `make html` inside the docs folder. The outputs will be in _build/html/
You can start with adding the function signatures and their descriptions in the docs/reference folder by copying an existing file and renaming it to io.rst.

Please let me know if you have any questions!

nakul-shahdadpuri:
@ziyaointl 
I have updated the docs for the added functions.
Let me know about any issues.
Thank you. :)

p5/io/http.py
```diff
@@ -15,6 +22,15 @@ def http_get(url):
 
 #synchronous
 def http_post(url,post_data):
+	"""
+	Performs a synchronous POST request on the given url.
+
+	:param url: url of the API
+	:type url: string
+	:param post_data: Key-value pairs of the request payload.
+	:type post_data: map
```
ziyaointl, line 31: Maybe this should be a `dict` since it's in Python?

ziyaointl:
Thanks for updating the docs! This is awsome.

p5/io/http.py
nakul-shahdadpuri: Ok I ll make the change. :)

PR #249, opened 2020-08-30
Add more sample scenes to profile

ziyaointl:
Adds more sample scenes in the `profiling` folder.
Also includes a simple bash harness to run all the scenes because the previous python one does not always work possibly due to some global state being preserved between scenes.


parsoyaarihant:
can you add a small readme file on how to run the profiling via the bash script?

ziyaointl:
As of now the bash script actually does not output any profile information - it just runs each scene for 3 seconds. While this is useful for integration testing, I think I'll need to combine it with a singleton Python harness that does the actual profiling in the future.

And yeah, I'll go ahead and add this information as well as how to use both in a README.

PR #256, opened 2020-08-30
Fixed Import Error and module collision.

nakul-shahdadpuri:
@tushar5526 @ziyaointl
 Regarding [issue](https://github.com/p5py/p5/issues/244), I think this should fix the issues mentioned.
Thank you for your feedback. :)


ziyaointl:
LGTM!

PR #257, opened 2020-09-01
Added load tables in IO module.

nakul-shahdadpuri:
@ziyaointl 
Regarding [issue](https://github.com/p5py/p5/issues/244),
I have added functions for loading csv files stored locally.
Please let me know about any changes needed.
Thank you. :)


tushar5526:
Any thoughts about using pandas, it will also provide you with extra functionalities? @ziyaointl  @nakul-shahdadpuri

nakul-shahdadpuri:
@tushar5526 @ziyaointl 
I was a little concerned regarding the overhead associated with pandas.
Also I am trying to keep the functions in sync with p5.js [tables](https://p5js.org/reference/#/p5.Table) module. 
Most of the functions in the module seemed easy enough to be implemented from scratch.

ziyaointl:
@tushar5526 @nakul-shahdadpuri I think using pandas is a good idea because it has a very robust parser that's difficult to replicate, but I also agree that keeping the API in sync with p5.js has its merits because some of the pandas API is arguably less user-friendly compared to those of p5.js.

This PR seems good as-is (after adding documentation), but before implementing more APIs for the Table class I think it's worth investigating if we can use pandas as the underlying data source and rewrite the curret API as a thin wrapper on top of that.

p5/io/local_data.py
```diff
@@ -0,0 +1,44 @@
+class table:
```
ziyaointl, line 1: nit: capitalize class names

p5/io/local_data.py
```diff
@@ -0,0 +1,44 @@
...
+
+	def __init__(self,PATH):
+		try:
+			file = open(PATH,'r')
+		except:
```
ziyaointl, line 6: I think it's ok to not catch the error here and simply let upstream code handle it.

p5/io/local_data.py
```diff
@@ -0,0 +1,44 @@
...
+					break
+				count = count + 1
+			column = []
+			for item in self.data:
+				column.append(item[count])
+			return column
+		except:
+			print("ERROR: column does not exist")
+
+	def get_row(self,inde):
```
ziyaointl, line 39: nit: typo?

p5/io/local_data.py
```diff
@@ -0,0 +1,44 @@
...
+		try:
+			for i in self.data[0]:
+				if i == name:
+					break
+				count = count + 1
+			column = []
+			for item in self.data:
+				column.append(item[count])
+			return column
+		except:
```
ziyaointl, line 36: I think it's better to just throw an error if the designated column is not found instead of printing to stdout, because it's harder for the user to realize & handle the latter. Usually if a column name is not found then there is a bug with user code, so if the user program crashes that's a clearer indicator compared to an output in stdout. If there isn't a bug with user code, using an error also allows the user to handle it themselves.

p5/io/local_data.py
```diff
@@ -24,17 +21,14 @@ def get_column_count(self):
...
+		for i in self.data[0]:
+			if i == name:
+				break
+			count = count + 1
+		column = []
+		for item in self.data:
+			column.append(item[count])
+		return column
 
 	def get_row(self,inde):
```
tushar5526: inde -> index ?

nakul-shahdadpuri: Just pushed that change in. :)

nakul-shahdadpuri:
Added Docs and made the changes according to the review.
Will look into the pandas wrapper. :)

parsoyaarihant:
I think there needs to be a public function `loadTable` which will then call `Table` class: https://p5js.org/reference/#/p5/loadTable

nakul-shahdadpuri:
@parsoyaarihant Added load_table function along with its doc.
Let me know about any changes.:)

p5/io/local_data.py
```diff
@@ -0,0 +1,76 @@
...
+			x = x[:-1]
+			fragment[len(fragment)-1] = x
+			data.append(fragment)
+		self.data = data
+
+	def get_row_count(self):
+		"""
+		Returns number of rows in the read CSV.
+		"""
+		return (len(self.data))
```
ziyaointl, line 26: nit: extra parenthesis? are you trying to return a tuple?

p5/io/local_data.py
```diff
@@ -0,0 +1,76 @@
...
+			if fragment[0] == index:
+				return fragment
+	def get_array(self):
+		"""
+		Returns the entire csv data as an multidimensional array.
+		"""
+		return self.data
+
+
+def load_table(PATH):
```
ziyaointl, line 68: nit: function arguments should be lowercase

p5/io/local_data.py
```diff
@@ -0,0 +1,76 @@
...
+		"""
+		Returns number of rows in the read CSV.
+		"""
+		return (len(self.data))
+
+	def get_column_count(self):
+		"""
+		Returns number of columns in the read CSV.
+		"""
+		return (len(self.data[0]))
```
ziyaointl, line 32: same as above

p5/io/local_data.py
```diff
@@ -0,0 +1,76 @@
+#for loading tables in csv format
+class Table:
+
+	def __init__(self,PATH):
```
ziyaointl, line 4: nit: lowercase for PATH

ziyaointl:
@nakul-shahdadpuri Thanks for the great work and timely fixes! Sorry to have bothered you with many stylistic suggestions. Hopefully we can automate these with CI in the near future.

nakul-shahdadpuri:
@ziyaointl My pleasure. :)

PR #264, opened 2020-09-03
Fixed get_row error and added ssv support.

nakul-shahdadpuri:
Regarding [issue](https://github.com/p5py/p5/issues/244),
@ziyaointl 
Added support for ssv files.
Fixed row_count() to be consistent with p5.js API
Updated docs to add return types.
Will push the pandas wrapper soon.


p5/io/local_data.py
```diff
@@ -52,27 +55,38 @@ def get_column(self,name):
...
 		"""
-		Returns the entire csv data as an multidimensional array.
+		:returns: the entire csv data as an multidimensional array.
+		:rtype: list
 		"""
 		return self.data
 
 
-def load_table(path):
+def load_table(path,mode):
```
tushar5526, line 73: p5.js has other optional arguments too like header, error, and success callbacks. You can use *args to take multiple optional variables, would-be future compatible. So something like 
```
def load_table(path, *args):
    table = Table(path, *args)
```

p5/io/local_data.py
```diff
@@ -52,27 +55,38 @@ def get_column(self,name):
...
+
+	:returns: A Table object
+	:rtype: object
 	"""
-	table = Table(path)
+	if mode == "csv":
+		seperator = ','
+	if mode == "ssv":
+		seperator = ';'
+
```
tushar5526, line 90: There is one more mode "tsv" "\t", add that too. :)

nakul-shahdadpuri: I ll add that too 😀.

parsoyaarihant:
We have recently used pylinter to format the code (#269) and added pylinter tests. Can you update your branch to master and add the changes again? 
Thanks.

nakul-shahdadpuri:
@parsoyaarihant  Resolved the conflicts. :)

ziyaointl:
Looks good to me. Thanks!

PR #276, opened 2020-09-15
Fixes

Andy-Python-Programmer:
I was unable to do this fix as I was buzy. Now I have done them.
```python
def setup():
    size(400, 400)

def draw():
    background(0)
    
    example = http_get("https://example.com")
    print(example.text)

run()
```
I have made it a class so that we can do example.text or example.json (for json). If we would have checked each of the lines for example decode json if it startswith("{") and endswith("}") then it will be a very slow process.

### DIFF
--- a/p5/io/http_data.py
+++ b/p5/io/http_data.py
@@ -18,18 +18,15 @@ class http_get:
     """
 
     def __init__(self, url):
-        try:
-            resp = request.urlopen(url)
-        except:
-            print("Invalid url. Connection Failed.")
-            exit(1)
+        resp = request.urlopen(url)
 
         data = resp.read().decode("UTF-8")
 
         self.text = data
+
         try:
             self.json = json.loads(data)
-        except: 
+        except json.JSONDecodeError:
             self.json = data
 
 # Synchronous
--- a/p5/io/http_data.py
+++ b/p5/io/http_data.py
@@ -1,11 +1,14 @@
-from urllib import request, parse
+"""
+Part of the io library and has functions that require **internet** connection.
+"""
+
+from urllib import request
 import json
 import requests
 
-# synchronous
+# Synchronous
 
-
-def http_get(url):
+class http_get:
     """
     Performs a synchronous GET request on the given url.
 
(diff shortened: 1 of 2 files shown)


p5/io/http_data.py
```diff
@@ -1,29 +1,38 @@
-from urllib import request, parse
+"""
+Part of the io library and has functions that require **internet** connection.
```
nakul-shahdadpuri, line 2: remove "and" from comments

p5/io/http_data.py
```diff
@@ -1,29 +1,38 @@
...
+from urllib import request
 import json
 import requests
 
-# synchronous
-
+# Synchronous
 
-def http_get(url):
+class http_get:
```
nakul-shahdadpuri, line 11: make the changes in docs accordingly.

p5/io/http_data.py
```diff
@@ -1,29 +1,38 @@
...
+            print("Invalid url. Connection Failed.")
+            exit(1)
+
+        data = resp.read().decode("UTF-8")
+
+        self.text = data
+        try:
+            self.json = json.loads(data)
+        except: 
+            print("The data provided did not return in a json format.")
```
nakul-shahdadpuri, line 33: return the data in string if not JSON.

nakul-shahdadpuri (changes requested):

parsoyaarihant:
can you be a bit more descriptive about this PR? 

It would be great if you could refer to the related issues and PRs.

Andy-Python-Programmer:
@parsoyaarihant So the issue was that the get request from now only returned json so it failed it the result is not in json format. This pr fixes that issue. No related issues are there.

parsoyaarihant:
I have few comments about this module: 

1. If the request failed, it should be more descriptive about why the request failed. it should give the status codes: https://www.restapitutorial.com/httpstatuscodes.html
2. If you are making a class based, implementation, all the functions should use that implementation. Not just http_get(). 
3. What about asynchronous calls? Should we pass a callback function for asynchronous requests? : https://p5js.org/reference/#/p5/httpGet

Andy-Python-Programmer:
@parsoyaarihant for the 2. Point. I think we only need class for http_get we do not need class based implementation for http_post. And I will take a look at the http error codes.

Andy-Python-Programmer:
Hi @parsoyaarihant about the error codes what we can do is just simply remove the try and except as URL lib will auto throw an exception and it will contain the respected http error codes.

p5/io/http_data.py
```diff
@@ -1,29 +1,38 @@
...
+        except:
+            print("Invalid url. Connection Failed.")
+            exit(1)
+
+        data = resp.read().decode("UTF-8")
+
+        self.text = data
+        try:
+            self.json = json.loads(data)
+        except:
```
parsoyaarihant, line 32: When do we expect `json.loads()` to fail?

parsoyaarihant:
We can do that, I think the documentation and examples should also contain the try and except statements to explain how to use this API

p5/io/http_data.py
Andy-Python-Programmer: Well if the resp is 
```dart
test
```
then fail or 
```json
{wrong syntax
```

parsoyaarihant:
Thanks

PR #279, opened 2020-09-17
added tests for storage module

tushar5526:
Added tests for storage module. 
Feedbacks are welcomed as I haven't wrote tests before.
Thanks :)

### DIFF
--- a/p5/data/tests/test_data.py
+++ b/p5/data/tests/test_data.py
@@ -1,0 +1,30 @@
+import unittest
+from p5.data import clear_storage, get_item, remove_item, set_item
+
+
+class TestStorage(unittest.TestCase):
+    def test_set_item(self):
+        set_item('test', 'p5py')
+        self.assertEqual(get_item('test'), 'p5py')
+
+    def test_get_item(self):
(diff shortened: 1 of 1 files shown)


parsoyaarihant:
Thanks

jeremydouglass:
@tushar5526 my one piece of feedback would be that these tests should better cover the expected limitations of the code.

-  If the user passes an empty string as their key, what happens, and is it what we expected?
-  If the user passes "/" as a key -- a forbidden file character in Windows and Linux/Unix -- what happens, and is it what we expected?
-  If the user tries to get an item that doesn't exist, what happens?
-  If the user clears storage, and then clears it again while storage is empty, what happens.

The answer might be "of course, nothing" but tests help with regressions -- so that if the system is changed in the future, assumptions are checked by testing. One of the main things I wonder when I look at this code and think about writing tests for it is, is it safe enough? This is because it is writing and deleting files whenever called. So, for example, set_item deletes the existing path with os.remove before writing it. If the user calls `set_item("../../", '')` will that delete the whole project directory? Or is it safe? Etc.

Issue #214, opened 2020-08-02
Update Public APIs to support different renderers


parsoyaarihant:
Currently all the APIs use PShape only for rendering shapes as vispy is the only renderer being used. To add support for new renderers, we want to update the existing API functions to call that specific renderer. Pseudo code:

``` python
def shapeXYZ(): # some rendering function
    renderer.shapeXYZ() # the function corresponding to the renderer will be called.
```

parsoyaarihant:
Few more things needs to be done after #288 

1. Implement begin_shape() API inside the Vispy2DRenderer

parsoyaarihant:
2. Test if the Font and Image APIs are working as expected with the new library structure.

tushar5526:
It skipped my mind to refactor `image.py` as it has a dependency on **Texture2D** from `vispy.gloo`. Image APIs are working fine as of now, as there is only one renderer. 
`PImage` has to be updated to support multiple renderers. I will make a PR soon.

parsoyaarihant:
#289

tushar5526:
> Few more things needs to be done after #288
> 
> 1. Implement begin_shape() API inside the Vispy2DRenderer

@parsoyaarihant  Can you tell a bit more about this? I am interested in doing it.

parsoyaarihant:
begin_shape() API currently uses PShape: https://github.com/p5py/p5/blob/master/p5/core/vertex.py#L335

We need to move this code into the VispyRenderer.

PR #288, opened 2020-10-17
Refactored Vispy2D renderer under one module 

tushar5526:
`Userspace.py` now only imports and uses vispy as renderer when selected.
`Base.py` and `renderer2D.py` is refactored.

Before working on integrating Cairo and refactoring code, should we work on writing **visual tests** for p5py?
As sooner or later we will need that and manually checking for bugs when we will start integrating Cairo will become hectic.
 
**This PR is not tested robustly, I just tested it on some 2D and 3D examples, and things were working fine :)**


p5/sketch/userspace.py
```diff
@@ -143,34 +139,35 @@ def run(sketch_setup=None, sketch_draw=None,
...
-    builtins.pixel_x_density = physical_width // width
-    builtins.pixel_y_density = physical_height // height
-    builtins.start_time = time.perf_counter()
+        p5.sketch.timer.start()
 
-    p5.sketch.timer.start()
-
-    app.run()
-    exit()
+        app.run()
```
parsoyaarihant, line 167: This line is visy specific. I believe if we use another renderer, it will throw an error.

parsoyaarihant:
To test the changes, we can directly checkout your branch locally and test it. We dont have to create a new branch in p5py.

parsoyaarihant:
Nice work! I will test this properly in the coming days. 

For #214 we now need two things:

1.  Refactor the PShape class into the newly created vispy renderer. Basically move the shape.py file into the Vispy2DRenderer folder.
2. Create new methods in Vispy2DRenderer class for the primitive APIs (like line, rectangle, etc). Then call these functions from the primitives.py. The primitive shape functions should not return a PShape object like the currently do, instead they should call the corresponding function in the Vispy2DRenderer.

parsoyaarihant:
Can you put `shaders2d.py` (and shader files) and `openglrenderer` inside the `Vispy2DRenderer`?

p5/sketch/userspace.py
tushar5526: `app.run()` will be executed only when the renderer is set to `vispy`. 
`app.run()` and all other `vispy` specific calls is under the `if renderer == "vispy"` clause. I think Github's UI is making it confusing to figure out the indents.

Or is it something else that I am missing, can you confirm this @parsoyaarihant ?
Thank you for the review :)

tushar5526:
> 1. Refactor the PShape class into the newly created vispy renderer. Basically move the shape.py file into the Vispy2DRenderer folder.
> 2. Create new methods in Vispy2DRenderer class for the primitive APIs (like line, rectangle, etc). Then call these functions from the primitives.py. The primitive shape functions should not return a PShape object like the currently do, instead they should call the corresponding function in the Vispy2DRenderer.

Thanks for the detailed description of what needs to be done. I will start working on it, but due to exams will be little unactive till the month's end.

tushar5526:
Now, `renderer2d` and `renderer3d` is imported after we check which *renderer mode and renderer backend* is specified by the user.
I have also encapsulated all 3d renderer code ( shader files, renderer3d.py) in Vispy3DRenderer, otherwise `renderer3d` was loading **shader files** before `builtins.current_renderer` was defined.
I tried to generalize things as much as possible so that it's easy to add other backends for both 2D and 3D in the future.

tushar5526:
I will be making PRs to this branch from now onwards or there will be too much code to review :)

.gitignore
```diff
@@ -111,3 +111,8 @@ ENV/
 
 # Profiling results
 profiling/*.prof
+
+.idea/
+test.py
```
parsoyaarihant, line 116: This is useful for development. But we need to discuss and document this properly.

p5/core/primitives.py
```diff
@@ -25,7 +25,7 @@
 from ..pmath import curves
 from ..pmath.utils import SINCOS
 
-from .shape import PShape
+from p5.sketch.Vispy2DRenderer.shape import PShape
```
parsoyaarihant, line 28: We dont have to use the entire namespace. We can use `from ..sketch.Vispy2DRenderer.shape import PShape`

tushar5526: These changes in namespaces were done automatically by **pycharm** when I moved (refactored) the files. Should I undo this manually ?

.gitignore
tushar5526: I have documented this.

tushar5526:
@parsoyaarihant before I redo the APIs, can you confirm is this the intended design we are looking for? 
```
def quad(*args):
    """doc string"""
    
    (parse args here)
    if len(args) == 8:
        p1, p2, p3, p4 = args[:2], args[2:4], args[4:6], args[6:]
    elif len(args) == 4:
        p1, p2, p3, p4 = args
    else:
        raise ValueError("Unexpected number of arguments passed to quad()")
    
    (pass onto renderer then)
    p5.renderer.quad(p1, p2, p3, p4)
```
**Quad method in VispyRenderer Class**

```
    def quad(self, *args):
        p1, p2, p3, p4 = args
        path = [
            Point(*p1),
            Point(*p2),
            Point(*p3),
            Point(*p4)
        ]
        self.render(PShape(vertices=path, shape_type=SType.QUADS))
```

ziyaointl:
Great work! Now that the shaders have moved, you may want to modify MANIFEST.in to include the new shader location.

parsoyaarihant:
> @parsoyaarihant before I redo the APIs, can you confirm is this the intended design we are looking for?

Yes, this is the design we are looking for

p5/core/primitives.py
parsoyaarihant: Fix these imports. Then I will merge this PR

parsoyaarihant:
> Before working on integrating Cairo and refactoring code, should we work on writing visual tests for p5py?

Cairo has anti-aliasing and Vispy does not. If we write the tests for Vispy, then they will fail with Cairo.

tushar5526:
I have compiled all the examples present at p5's read the docs for testing purposes. In case you need it, they are under **check_tests** at this [branch](https://github.com/tushar5526/p5/tree/p5Examples)

PR #289, opened 2020-11-01
Refactored image.py to support different renderers

tushar5526:
`_texture` property now returns the data of the image instead of a `gloo.Texture2D` object. Texture objects are now created in the renderer itself.

### DIFF
--- a/p5/core/image.py
+++ b/p5/core/image.py
@@ -29,7 +29,6 @@ from PIL import Image
 from PIL import ImageFilter
 from PIL import ImageChops
 from PIL import ImageOps
-from vispy import gloo
 from . import p5
 
 from . import color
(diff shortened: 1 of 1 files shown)


ziyaointl:
It seems that you removed the caching built into `_texture`. Have you run some profiling to see the performance impact of this?
Otherwise looks great!

tushar5526:
> It seems that you removed the caching built into `_texture`. Have you run some profiling to see the performance impact of this?

This may lead to some overheads, I will check it.
Thanks!

tushar5526:
Thanks @ziyaointl I overlooked the caching part which led to significant overhead.
I have made some changes. Please take a look.

ziyaointl:
Thanks for the changes!

parsoyaarihant:
#214

PR #291, opened 2020-11-10
Refactor begin_shape & end_shape APIs to support multiple renderers

tushar5526:
Related to #214 
`end_shape` now calls `p5.renderer.shape()` and passes the required data instead of returning a `PShape` obejct and using `@draw_on_return`

### DIFF
--- a/p5/core/primitives.py
+++ b/p5/core/primitives.py
@@ -16,38 +16,22 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 
-import functools
 import math
 
 from ..pmath import Point
 from ..pmath import curves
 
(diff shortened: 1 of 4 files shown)


parsoyaarihant:
We can also remove `@draw_on_return` from primitives.py as its no longer needed: https://github.com/p5py/p5/blob/master/p5/core/primitives.py#L39

tushar5526:
@parsoyaarihant 
https://github.com/p5py/p5/blob/master/p5/core/primitives3d.py#L49
There are still some dependencies on `@draw_on_return`. Should I move `@draw_on_return` to `primitves3d.py` ?

parsoyaarihant:
Yes.

parsoyaarihant:
Thanks

PR #297, opened 2020-12-03
bug fix : support new APIs to renderer PShape in 3D renderer as well

### LINKED ISSUE #296: Vertices when called in 'P3D' throws error.
`vertex.py` calls `p5.renderer.shape()` ( the new 2D renderer APIs ) but when used in 'P3D' mode it throws an error as the 3D renderer does not have a `shape()` function.

```

from p5 import *

def setup():
    size(640, 360)
    no_loop()

def draw():
    background(0)

    stroke(255)
    rotate_x(PI*3/2)
    rotate_z(PI/6)
    fill(255)
    begin_shape()
    vertex(-100, -100, -100)
    vertex( 100, -100, -100)
    vertex(   0,    0,  100)

    vertex( 100, -100, -100)
    vertex( 100,  100, -100)
    vertex(   0,    0,  100)

    vertex( 100, 100, -100)
    vertex(-100, 100, -100)
    vertex(   0,   0,  100)

    vertex(-100,  100, -100)
    vertex(-100, -100, -100)
    vertex(   0,    0,  100)
    end_shape()

run(mode='P3D')
```

tushar5526:
Closes #296 and other minor updates. 
`push()` was added to Processing that saves current style as well as transformations.

### DIFF
--- a/p5/core/api.py
+++ b/p5/core/api.py
@@ -14,8 +14,16 @@ from .vertex import begin_shape, begin_contour, curve_vertex, bezier_vertex, end
 from .material import normal_material, basic_material, blinn_phong_material
 from .light import ambient_light, directional_light, point_light, light_falloff, light_specular
 
-
 def push():
+    """	The push() function saves the current drawing style settings and transformations"""
+    push_matrix()
+    push_style()
+
+def pop():
+    pop_style()
+    pop_matrix()
+
+def pushStyle():
     """Save the current style settings and then restores them on exit.
 
     The 'style' information consists of all the parameters controlled
@@ -49,7 +57,7 @@ def push():
     push_style()
 
 
-def pop():
+def popStyle():
     """Restores previously pushed style settings
     """
     pop_style()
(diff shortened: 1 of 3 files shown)


p5/core/api.py
```diff
@@ -14,8 +14,16 @@
 from .material import normal_material, basic_material, blinn_phong_material
 from .light import ambient_light, directional_light, point_light, light_falloff, light_specular
 
-
 def push():
+    """	The push() function saves the current drawing style settings and transformations"""
```
parsoyaarihant, line 18: wasnt this there before?

tushar5526: I wrongly named `pushStyle` as `push` back then when I added the support for processing APIs.

Here is the [link](https://processing.org/reference/push_.html) for the `push()` API that was missing in p5py.
There is `push_matrix` and `push_style` present in p5py but no `push()` that stores both **transformation** and **style** changes.

`push` in p5.js [link](https://p5js.org/reference/#/p5/push)

Found this error after trying to implement this example.

```
fill(255);
rect(0, 0, 50, 50);  // White rectangle

push();
translate(30, 20);
fill(0);  
rect(0, 0, 50, 50);  // Black rectangle
pop();  // Restore original settings

fill(100);  
rect(15, 10, 50, 50);  // Gray rectangle
example pic
ellipse(0, 50, 33, 33);  // Left circle

push(); 
strokeWeight(10);
fill(204, 153, 0);
ellipse(50, 50, 33, 33);  // Middle circle
pop();  // Restore original settings

ellipse(100, 50, 33, 33);  // Right circle
```
The results were different in p5py and p5.js.

PR #309, opened 2021-04-25
Fix exceptions shown when closing canvas

### LINKED ISSUE #150: AttributeError: module 'p5.core.p5' has no attribute 'exit'
**Describe the bug**
I tried running the sample code in p5 README text in spyder 4, Jupyter Notebook, and external windows terminal. The same error(AttribureError: ...) pops up in the first two. The terminal , however properly executes this code. Please let me know how to fix it. (i find it comfortable to run codes in those environments...)
**To Reproduce**
Try running the code in those environments:

[samplep5.txt](https://github.com/p5py/p5/files/4412925/samplep5.txt)

**Screenshots**
The errors i got in the two environments
[error_jupy.txt](https://github.com/p5py/p5/files/4412920/error_jupy.txt)
:
[error_spy.txt](https://github.com/p5py/p5/files/4412916/error_spy.txt)

**System information:**
 - p5 release 0.6.0
 - Python 3.7.6
 - Operating system: Windows 10

andreasWallner:
Fixes #150 and additionally prevents a SystemExit exception being thrown/reported that is anyways ignored.

`p5.exit()` documentation mentions that it forwards args
to the builtin `exit()`, so do that instead of calling
the non-existing `p5.core.p5.exit()` function. (Or prevent
a recursive call that by exit ended up calling a
non-existant function.

Do not call `p5.exit()` in on_close: exit() will throw a SystemExit
exception which will be cought by vispy, reported and ignored.
Since the windows has gone away we only need to quit the
running application for a normal exit.

### DIFF
--- a/p5/sketch/Vispy2DRenderer/base.py
+++ b/p5/sketch/Vispy2DRenderer/base.py
@@ -147,7 +147,7 @@ class VispySketch(app.Canvas):
         self._save_flag = True
 
     def on_close(self, event):
-        exit()
+        app.quit()
 
     def on_draw(self, event):
         pass
--- a/p5/sketch/userspace.py
+++ b/p5/sketch/userspace.py
@@ -22,6 +22,7 @@ import __main__
 import math
 import numpy as np
 import builtins
+import sys
 import time
 from functools import wraps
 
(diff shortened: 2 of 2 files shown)


ziyaointl:
I was not able to reproduce the original error, but the changes included in this PR is definitely more logical than the original code. Thank you @andreasWallner!

PR #310, opened 2021-04-29
make save() consistent with processing

tushar5526:
There is no need to add `_save_fname_num` to the image name. Images are saved by the default name or by the `filename` passed.


ziyaointl:
Looks good. Thanks @tushar5526!

PR #311, opened 2021-05-28
Convert 2D vertices to 3D for transformation, issue 199

yogeshhk:


ziyaointl:
Thanks! This looks great. I also tested it locally on the sample scenes and they all worked.

PR #308, opened 2021-04-08
Fix vertex transformation in 2D renderer

janbehrens:
This is how I resolved #307. Let me know if it makes any sense.

Issue #89, opened 2019-01-15
p5 doesn't exit if there is a bug in draw or setup.


redfood:
**Describe the bug**
If you have an error in your program you would expect for python to print the error and to exit.  Instead, if you have an error in `draw` or `setup` p5 doesn't exit. It prints the error and then repeatedly prints `ERROR: Invoking <bound method Sketch.on_timer of <Sketch (Glfw) at 0x10fb3ed68>> for Event`.  This is true even if you have called `no_loop()`

**To Reproduce**
Introduce an error into `setup` or `draw`.  e.g.
``` py
from p5 import *

def setup():
    size(100,100)

def draw():
    no_loop()
    typo()

run()
```

**Expected behavior**
You would expect the program to exit with an error message.

**System information:**
 - p5 release (version number or latest commit): 7.2
 - Python version: 3.7.1
 - Operating system: macOS 10,13.6

dbhavsar76:
I think this is GLFW or Vispy's problem.

abhikpal:
> I think this is GLFW or Vispy's problem.

Correct. I can't think of any way we can handle this from within p5 as everything inside draw/setup is ultimately sent to vispy (and hence GLFW). I'm closing this now, but if someone has ideas, feel free to open this.

PR #318, opened 2021-08-16
Basic sanity testing for 2D and 3D sketches.

tushar5526:
To help the maintainers review new PRs easily, I have put together some p5 sketches from the documentation covering various APIs of the p5py package. 

It runs all the sketches for 2 seconds headlessly and checks for errors and prevents PRs from being merged that would break the functionality of the library. 

#308 broke the 3D rendering pipeline, so I have reverted the changes. Maybe @janbehrens @ziyaointl could look into it.


ziyaointl:
Thank you! This is excellent work and I'm sure all p5py contributors would appreciate this addition to the project. I'll take a closer look at the renormalization problem in the 3D pipeline today and post an update :)

p5/visualTests/Instructions.md
```diff
@@ -0,0 +1,12 @@
+# Run the visual test
+```
+python sanity_testing.py
+```
+
+**NOTE**: If you get a `ModuleNotFound or ImportError` make sure you export the `project's root path` to `PYTHONPATH` variable.
+
+```
+export PYTHONPATH=$PYTHONPATH:path/to/project"
+```
```
tushar5526, line 10: @ziyaointl  Are you familiar with any other way to export the project's root path to the PYTHONPATH environment variable?
I was getting a ModuleNotFound error on running scripts through the terminal, and scripts were running fine if I was using the IDE's command to run them `Ctrl + shift + F10` in pycharm as the IDE automatically adds the project's root path to the variable.

.github/workflows/sanityTesting.yml
```diff
@@ -0,0 +1,31 @@
...
+          python-version: 3.8
+      - name: Install dependencies
+        run: |
+          sudo apt-get install python-opengl libglfw3 python3-wheel -y
+          python -m pip install --upgrade pip
+          pip install wheel
+          pip install -r requirements.txt
+      - name: Set env
+        # TODO: Find an alternate way to get the path, instead of hard-coding
+        run: echo "PYTHONPATH=$PYTHONPATH:${HOME}/work/p5" >> $GITHUB_ENV
```
tushar5526, line 25: I have hardcoded this path, but it will break if GitHub will change its configuration of running actions in a different folder inside the VM. We can keep this for now, but it is a potential bug.

p5/visualTests/sanity_testing.py
```diff
@@ -0,0 +1,41 @@
+import subprocess
+import os
+from subprocess import run, PIPE
+
+e = SystemExit()
+
+dirname = os.path.dirname(__file__)
```
ziyaointl, line 7: I think `os.path.dirname(os.path.realpath(__file__))` is more robust, since it handles relative directories as well.
Source: https://stackoverflow.com/questions/5137497/find-current-directory-and-files-directory

.github/workflows/sanityTesting.yml
ziyaointl: Similarly, I think `pip install .` could also work here.

p5/visualTests/Instructions.md
ziyaointl: I prefer `pip install .` if I'm in a virtual env since it's a lot easier to type and it isn't a big deal to mess up the installed packages. Furthermore, this makes sure that an pre-existing p5 package won't interfere with the new changes we are trying to test. Maybe we should use this instead?

p5/visualTests/sanity_testing.py
tushar5526: Great, didn't know that and it was breaking when I ran sanity_testing from its directory. Thanks

p5/visualTests/Instructions.md
tushar5526: Worked. Thanks!

.github/workflows/sanityTesting.yml
tushar5526: Worked perfectly. Thanks!

tushar5526:
Just to mention, we are setting up p5 multiple times for each of our actions. Instead, we can save and upload it as an artefact and save some processing power. I will cover this in a different PR soon.

ziyaointl:
> Just to mention, we are setting up p5 multiple times for each of our actions. Instead, we can save and upload it as an artefact and save some processing power. I will cover this in a different PR soon.

That's a good idea! Thanks for working on it!

Also, I found the reason of why the 3D pipeline was breaking - #308 did not convert the return value to an `np.array`. I'll open another PR that fixes the issue after this one is merged :)

p5/visualTests/sanity_testing.py
```diff
@@ -0,0 +1,39 @@
...
+
+print('============================= Running 2D tests =============================')
+for name in test_names_2d:
+    print('Running', name)
+    try:
+        # Run each test for 2s then raise a timeout
+        process = run(['python', name], stderr=PIPE, encoding='utf-8', timeout=2)
+    except subprocess.TimeoutExpired:
+        pass
+    else:
```
ziyaointl, line 23: I tested this locally but wasn't able to get the program to error even if I tried to manually break a function in p5. I think this is because vispy is capturing the exception is internally. As to why there is also no output to stderr - I'm not sure.

This said, I found a way to disable this exception capturing in vispy: add `self.events.ignore_callback_errors = False` to Vispy2DRenderer/base.py:69. Then, you should be able to just raise any exception here that's not `subprocess.TimeoutExpired`.

tushar5526: After adding some breaking changes, did you reinstalled the project again ? `pip install .`
The tests are referring to the installed p5 package instead of the current contents in project. Maybe that was the reason you could not get it to error. Can you confirm this ? 
To use the current project, without reinstalling again, we can export the project's path to PYTHONPATH locally. 
Thanks.

.github/workflows/sanityTesting.yml
```diff
@@ -0,0 +1,31 @@
...
+          sudo apt-get install python-opengl libglfw3 python3-wheel -y
+          python -m pip install --upgrade pip
+          pip install wheel
+          pip install .
+#      - name: Set env
+#        # TODO: Find an alternate way to get the path, instead of hard-coding
+#        run: echo "PYTHONPATH=$PYTHONPATH:${HOME}/work/p5" >> $GITHUB_ENV
+#      - name: Test
+#        run: echo $PYTHONPATH
+      - name: Setup virtual frame buffer
```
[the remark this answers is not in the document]
tushar5526: Sure.

p5/visualTests/sanity_testing.py
tushar5526: Its weird, vispy's window doesn't terminates after raising an error. I tried `self.events.ignore_callback_errors = False`, but it is not showing any difference. It logs there is an error but doesn't exits with a non-zero exit code.
Also found #89 related to it.

ziyaointl: Yes, I confirm that I did `pip install .` after every change. I went back and tested `self.events.ignore_callback_errors = False` again and it seems that it is not the only `EmitterGroup` that may be intercepting exceptions - there is another one in the `timer`. Could you try adding `self.timer.events.ignore_callback_errors = False` to `VispySketch.__init__` as well?

tushar5526: Works!!!
Thanks a ton @ziyaointl 
Any chances there might be other emitter groups, we have to disable callbacks ?

ziyaointl: You are welcome! A quick search indicates that the vispy `Visual` may also contain an `EmitterGroup`, but not sure if p5py uses that at all. https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/vispy/vispy%24+EmitterGroup&patternType=literal
I think it could be sufficient for us to turn more of this flag off if we encounter more unraised Exceptions later.

tushar5526: Great! This also seems to be the solution for #89

p5/visualTests/sanity_testing.py
```diff
@@ -0,0 +1,39 @@
...
+
+print('============================= Running 3D tests =============================')
+for name in test_names_3d:
+    print('Running', name)
+    try:
+        # Run each test for 2s then raise a timeout
+        process = run(['python', name], stderr=PIPE, encoding='utf-8', timeout=2)
+    except subprocess.TimeoutExpired:
+        pass
+    else:
```
ziyaointl, line 36: One last detail: I think we'll expose more exception details if we delete the else clause here.

ziyaointl: Oops I forgot that you are running it in a subprocess - yeah we should keep this else clause.

ziyaointl:
Thank you @tushar5526 for your timely responses! I'll go ahead and merge this PR :)

PR #322, opened 2021-08-21
Update minimum python version in tests

ziyaointl:
Thanks to @jeremydouglass who spotted that the automated tests use a python version higher than the minimally supported one, this PR updates the tests to use Python 3.6.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -9,10 +9,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.8
+      - name: Set up Python
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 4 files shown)


ziyaointl:
Also remove numpy from `setup_requires` since the newest version of numpy does not support python 3.6 and removing it didn't break setup.py.

tushar5526:
All the tests ran twice. I don't think it is the intended behaviour?

ziyaointl:
> All the tests ran twice. I don't think it is the intended behaviour?
> They ran once on PRs from forked repositories.

I think this is probably because I based this branch off of the old `master`, without the fix for the tests running twice.
If you open a new PR with the newest `master` as base, this problem should be fixed :)

PR #327, opened 2021-09-13
added badges for workflows

tushar5526:
Test status badges 🚀


ziyaointl:
Thank you! This looks great!
One more thing: maybe you should change the links to point to `p5py/p5` instead of `tushar5526/p5`?

tushar5526:
I missed that, thanks for catching that :)

ziyaointl:
np!

PR #344, opened 2022-06-11
Integrating skia

tushar5526:
Set up a basic rendering window using GLFW and skia, which supports the following APIs.

## Environment
- [x] frameCount
- [ ] deltaTime
- [x] Focused
- [ ] cursor()
- [x] frameRate()
- [ ] noCursor()
- [ ] displayWidth
- [ ] displayHeight
- [ ] windowWidth
- [ ] windowHeight
- [ ] windowResized()
- [x] Width
- [x] Height
- [ ] fullscreen()
- [ ] pixelDensity()
- [ ] displayDensity()	
## Structure
- [x] setup()
- [x] draw()
- [x] noLoop()
- [x] loop()
- [x] isLooping()
- [ ] push()
- [ ] pop()
- [x] redraw()

### Keyboard
- [x] keyIsPressed
- [x] Key
- [ ] keyCode (not used in p5py)
- [x] keyPressed()
- [x] keyReleased()
- [ ] keyTyped() (needs to be fixed on vispy end)
- [ ] keyIsDown() (not in p5py as of now, we can achieven this using key_is_pressed, so no need to implement it right now)
### Mouse
- [x] movedX
- [x] movedY
- [x] mouseX
- [x] mouseY
- [x] pmouseX
- [x] pmouseY
- [ ] winMouseX (no direct API in glfw, hacks might not work on all OS)
- [ ] winMouseY (no direct API in glfw, hacks might not work on all OS)
- [ ] pwinMouseY (no direct API in glfw, hacks might not work on all OS)
- [ ] pwinMouseX (no direct API in glfw, hacks might not work on all OS)
- [x] mouseButton
- [x] mouseIsPressed
- [x] mouseMoved()
- [x] mouseDragged()
- [x] mousePressed()
- [x] mouseReleased()
- [x] mouseClicked()
- [x] doubleClicked()
- [x] mouseWheel()

## Set up the rendering APIs for basic primitive shapes and set up the event handlers using glfw.
2D Primitives

- [x] arc()
- [x] ellipse()
- [x] circle()
- [x] line()
- [x] point()
- [x] quad()
- [x] rect()
- [x] square()
- [x] triangle()


p5/sketch/Skia2DRenderer/handlers.py
```diff
@@ -0,0 +1,67 @@
...
+
+# TODO : Make a raw_event class and populate it with data about current events
+# TODO : Design of raw_event should follow that of Vispy's raw event beacuse events.py is highly based on it  
+
+BUTTONMAP = {glfw.MOUSE_BUTTON_LEFT: 1,
+                glfw.MOUSE_BUTTON_RIGHT: 2,
+                glfw.MOUSE_BUTTON_MIDDLE: 3
+                }
+
+class PseudoMouseEvent:
```
tushar5526, line 19: @ziyaointl I am using a `PseudoEventClass` to mimic the data that was generated and need while using vispy-based events because the events.py is highly dependent on the `event` object passed by vispy. 

Please share your thoughts on this.

mrdaybird:
Checkout moderngl-window for how they implement event handling: https://github.com/moderngl/moderngl-window/tree/master/moderngl_window/context/glfw

p5/sketch/Skia2DRenderer/handlers.py
ziyaointl: I think this is solution has the pro of modifying minimal code in the current codebase, but I think in terms of keeping abstractions between renderers clean we may want to not include information about vispy inside of the skia renderer.

One solution I propose is to first modify the `Event`, `MouseEvent`, and `KeyEvent` classes in p5py so they don't directly rely on the raw events emitted by vispy, but rather specific information relavant from those events (e.g. `event.pos, event.buttons`). Then, one could write two helper functions that translate either vispy events or glfw events to p5py events. This seems to be similar to what @mrdaybird wrote in his proposal.

Let me your thoughts! If you also think this is a good idea, you can either create a new PR that refactors the existing event classes and rebase this skia PR on top of it, or you can keep the `PseudoMouseEvent` for now and create the refactor PR later down the road :)

tushar5526: Apologies for the late response, I was shifting back home for summer break. 

Yes, this is a good idea. We should refactor the event classes. I will add a helper function in `util` to generate a p5py event according to the `builtins.current_renderer` we are using.  

I will keep the `PseudoMouseEvent` now as it is working fine now and I could quickly move to the renderer implementation. 

Also, the Renderer and Events modules are completely independent of each other so I am thinking of refactoring the events module later.

tushar5526:
Also, taking a look at `moderngl-window`, it seems like a great choice for us to later move on to. It would provide support for multiple `windows(glfw, pyglet, pyqt)` 

Though it's a bit late for me to redo everything now, but we can explore migrating to it later for implementing other Processing APIs easily.

p5/sketch/Skia2DRenderer/handlers.py
ziyaointl: No worries at all! I think putting helper functions in either `events.py` or the respective renderers may be slightly better than putting them into the global util file, but that's just my opionon. The rest of the plan sounds great. Thank you!

ziyaointl:
@tushar5526 On preprocessing inputs before passing them into the renderer, I think it make a lot of sense to use `if-else` for now. After we figure out what different preprocessing needs to be done for each renderer, we can potentially tweak the renderer interface and try to move some of the renderer specific logic to each renderer class. But before we actually implement the preprocessing code for skia, it would be pretty hard to know what API tweaks we should make.

Thanks for posting the video! This is incredible progress! It's amazing that there is no aliasing at all. I look forward to the day that we can set skia as the default renderer.

tushar5526:
@ziyaointl should I wait for you to review and merge this PR, before continuing with my next milestone?

I was thinking of making a branch from this one, and working parallelly on my next milestone.

ziyaointl:
@tushar5526 Thanks for asking! Feel free to continue with another branch, and we can rebase as needed.

p5/core/primitives.py
```diff
@@ -558,23 +583,32 @@ def arc(*args, mode=None, ellipse_mode=None):
...
+            dim = Point(width, height)
+        else:
+            raise ValueError("Unknown arc mode {}".format(emode))
+        p5.renderer.arc(center, dim, start_angle, stop_angle, mode)
+
+    elif builtins.current_renderer == 'skia':
+        if not should_draw() or start_angle == stop_angle:
+            return
+        x, y = coordinate[0], coordinate[1]
+        vals = mode_adjust(x, y, width, height, emode)
```
ziyaointl, line 609: It seems that the documentation for `arc` only has `{'CORNER', 'CENTER', 'RADIUS'}`, but `module_adjust` also supports `CORNERS` as a potential input. Should we sanitize the input here, or is the documentation missing a mode?

p5/core/primitives.py
```diff
@@ -477,19 +498,24 @@ def square(*args, mode=None):
...
-        mode = _rect_mode
+        if mode is None:
+            mode = p5.renderer.style.rect_mode
+
+        if mode == 'CORNERS':
+            raise ValueError("Cannot draw square with {} mode".format(mode))
+        return rect(coordinate, side_length, side_length, mode=mode)
+    elif builtins.current_renderer == 'skia':
+        if should_draw():
+            rect(*args)
```
ziyaointl, line 517: I think `rect` expects there to be at least 4 input parameters, whereas `square` has only 2 or 3?
https://p5.readthedocs.io/en/latest/reference/shape.html?#square

p5/core/transforms.py
```diff
@@ -176,20 +159,19 @@ def apply_matrix(transform_matrix):
...
     """
-    p5.renderer.transform_matrix = np.identity(4)
+    p5.renderer.reset_matrix()
 
 
 def print_matrix():
     """Print the transform matrix being used by the sketch.
     """
-    print(p5.renderer.transform_matrix)
+    print(p5.renderer.print_matrix())
```
ziyaointl, line 174: I think it's a little odd for `renderer.print_matrix` to not print the matrix but return a string instead. I think we could either make the renderer function print directly, or rename the renderer function.

p5/visualTests/sanityTests/2DSanityTests/test23_skia.py
```diff
@@ -0,0 +1,34 @@
...
+
+        mx = constrain(mx, inner, width - inner)
+        my = constrain(my, inner, height - inner)
+        fill(76)
+        rect(edge, edge, width - 2*edge, height - 2*edge)
+        fill(255)
+        ellipse([mx, my], radius, radius)
+
+if __name__ == '__main__':
+        run(renderer='skia')
```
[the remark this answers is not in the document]
tushar5526: Yes, we should do that. Earlier I considered calling both the backend renderers one by one, something like
```
if __name__ == '__main__':
    run(renderer='skia')    
    run()
```
But, this was breaking in tests with `noLoop` enabled, because we are running the test for `100` frames and then calling  `exit()`.

I think it would be easier your way. I will make the updates to the testing suite.

p5/sketch/userspace.py
```diff
@@ -167,6 +191,19 @@ def run(sketch_setup=None, sketch_draw=None,
 
         app.run()
         exit()
+    elif renderer == "skia":
+        from p5.sketch.Skia2DRenderer.base import SkiaSketch
+        from p5.sketch.Skia2DRenderer.renderer2d import SkiaRenderer
+
+        builtins.current_renderer = renderer
+        # import sketch here
```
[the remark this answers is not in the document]
tushar5526: Yes

p5/core/color.py
```diff
@@ -54,17 +53,14 @@ def color_mode(mode, max_1=255, max_2=None, max_3=None, max_alpha=255):
...
-
     if max_2 is None:
         max_2 = max_1
 
     if max_3 is None:
         max_3 = max_1
 
-    color_range = (max_1, max_2, max_3, max_alpha)
-    color_parse_mode = mode
+    p5.renderer.style.color_range = (max_1, max_2, max_3, max_alpha)
```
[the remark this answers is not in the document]
tushar5526: Yes, I will try to break my PRs even further next time.

p5/core/primitives.py
tushar5526: The documentation is missing a mode. 
[ellipseMode](https://p5js.org/reference/#/p5/ellipseMode
)
[rectMode](https://p5js.org/reference/#/p5/rectMode
)

p5/core/transforms.py
tushar5526: Fixed

p5/core/primitives.py
tushar5526: Yes, fixed. Thanks

p5/visualTests/sanityTests/2DSanityTests/test23_skia.py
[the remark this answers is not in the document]
tushar5526: I have updated the testing suite, and the GHA for sanity tests will fail for the skia renderer now as we have not implemented some APIs in skia in this PR. We could either not run tests for skia.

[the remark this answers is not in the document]
ziyaointl: Yeah, I think not running tests for skia is a reasonable approach for now.

p5/sketch/Vispy2DRenderer/renderer2d.py
```diff
@@ -375,3 +376,52 @@ def arc(self, *args):
     def shape(self, vertices, contours, shape_type, *args):
         """Render a Pshape"""
         self.render_shape(PShape(vertices=vertices, contours=contours, shape_type=shape_type))
+
+    def reset_transforms(self):
```
ziyaointl, line 380: Looks like this (along with a few other functions) is duplicated between vispy2d and 3d as well. Could you also move them to the common opengl renderer?

ziyaointl:
@tushar5526 Thanks for such a quick response! I added one more small comment, but otherwise I think this PR is in great shape and should be merged :D

p5/sketch/Vispy2DRenderer/renderer2d.py
tushar5526: Moved the functions, and declared an abstract `render` function in `openglrenderer` for `shape`

p5/sketch/Vispy2DRenderer/renderer2d.py
```diff
@@ -372,56 +371,3 @@ def arc(self, *args):
         mode = args[4]
 
         self.render_shape(Arc(center, dim, start_angle, stop_angle, mode))
-
-    def shape(self, vertices, contours, shape_type, *args):
-        """Render a Pshape"""
-        self.render_shape(PShape(vertices=vertices, contours=contours, shape_type=shape_type))
```
ziyaointl, line 377: The 2D renderer calls `render_shape` instead of `render` because `shape.children` needs to be separately rendered. Maybe it's ok to leave `shape` in separate renderers with separate implementation for now?

tushar5526: Yes, nice catch. Thanks

Issue #352, opened 2022-07-07
Gracefully exit sketches


tushar5526:
exit() should not stop the whole process. instead, just stop and exit from the main loop and continue execution of the python file after `run()`

exit() in p5py calls the `sys.exit` which halts the execution, instead, it should just stop the main drawing loop.

This would help when p5py is used as a part of some project or visualization.

Current behaviour
```
run() (user closes window or presses escape)
# more python code, that will never run
```

Expected behaviour
```
run() ( user closes window or presses escape)
# main loop stopped and execution continues from here
```

tushar5526:
Closed by #355

PR #355, opened 2022-07-11
Exiting gracefully

niklasr22:
This is how i handled #352


tushar5526:
Thanks for the contribution @niklasr22

I have tried removing the `sys.exit` from vispy renderer, but you will end up with some "None Object" error from vispy. See the failed tests for log.

We probably have to gracefully quit the vispy, disconnect timers etc.

p5/sketch/Vispy2DRenderer/base.py
```diff
@@ -93,32 +96,39 @@ def on_timer(self, event):
...
-
-            while len(self.handler_queue) != 0:
-                function, event = self.handler_queue.pop(0)
-                event._update_builtins()
-                function(event)
+                elif not self.looping:
+                    pass
+
+                if self.exiting:
+                    raise ExitException()
```
tushar5526, line 123: Why are we raising an exit exception here, instead of directly calling `self.close()` and returning.

p5/sketch/userspace.py
```diff
@@ -283,19 +283,15 @@ def exit(*args, **kwargs):
     that necessary cleanup steps are performed before exiting the
     sketch.
 
-    :param args: positional argumets to pass to Python's builtin
+    :param args: positional arguments to pass to Python's builtin
```
tushar5526, line 286: Docstring should be changed now, given we are now not calling the python's builtin function.

tushar5526:
Thank you for the prompt responses @niklasr22 

There is one issue though, we are now shadowing python's builtin `exit` function. So, if someone has to call the python's `exit` they would have to use `sys.exit`

I will wait for @ziyaointl thoughts on this, apart from this PR looks good, I have tested it locally.

p5/sketch/Vispy2DRenderer/base.py
```diff
@@ -86,6 +86,8 @@ def __init__(self, setup_method, draw_method,
         self._save_fname = 'screen'
         self._save_flag = False
 
+        builtins.frame_count = 0
```
ziyaointl, line 89: I'm not sure if this is necessary, since `frame_count` is also initialized here: https://github.com/p5py/p5/blob/69000548faf556a77e1cdad7f6633f2cdd47006a/p5/sketch/userspace.py#L58

p5/sketch/userspace.py
```diff
@@ -283,19 +283,13 @@ def exit(*args, **kwargs):
     that necessary cleanup steps are performed before exiting the
     sketch.
 
-    :param args: positional argumets to pass to Python's builtin
-        `exit()` function.
+    :param args: positional arguments are ignored.
```
ziyaointl, line 286: We can probably remove both `*args` and `**kwargs`, since we are not calling the python native `exit` anyway, and Processing's Java API doesn't take any arguments for the `exit` function either: https://processing.org/reference/exit_.html

ziyaointl (commented):
I tested locally and this works very well! Closing everything down properly is quite tricky, so this is quite impressive. Just two minor comments and this PR should be good to go.

p5/sketch/Vispy2DRenderer/base.py
tushar5526: If there would be two `run` calls in the sketch , then the second run call will take the old `frame_count` values, so this has to be reset.

@niklasr22 instead of resetting it to `0` it should be reset to `-1`

ziyaointl: Ah I see, thanks for pointing this out!

Issue #298, opened 2020-12-11
particle system run very slowly


yonghuming:
**Describe the bug**
simple particle system run slowly
```
from p5 import *
from random import randint

class Ball():
    def __init__(self):
        self.x = randint(0, 800)
        self.y = 0
        self.vy = randint(1, 5)
    
    def update(self):
        self.y += self.vy

    def display(self):
        ellipse(self.x, self.y, 100, 100)

    def run(self):
        self.update()
        self.display()
        
balls = []


def setup():
    global balls
    size(800, 600)
    balls = [Ball() for i in range(100)]
    

def draw():
    global balls
    [ball.run() for ball in balls]
    background('red')

run()
```

**To Reproduce**
just run the code above

**Expected behavior**
as least fluently like https://editor.p5js.org/yonghuming/full/BDkk1FPlP

**System information:**
 - p5 release (version number or latest commit): newest
 - Python version:3.7
 - Operating system: windows 10

**Additional context**

aaronpenne:
+1 seeing same behavior on 2017 MacBook Pro

ReneTC:
I think this issue will be solved once skia is implimented #212

yonghuming:
> skia

opengl is powerful, why p5py so slow? pygame have better experence with similar code, also arcade. mabye there are ways to figure it still use opengl and glfw

yonghuming:
skia use glfw

tushar5526:
You can try the new experimental renderer or visit #357. I tried the above example and it runs fine for skia

Issue #131, opened 2019-10-02
Sketch does not cover windows on retina display


swirly:
On a retina display, the skecth is only covering a quarter of the window. This is a reopening of #9 since the supposed fix does not seem to work

In order to reproduce, use a retina display and just launch a script. It should cover the entire windows, it does not

Here is a simple screenshot
<img width="721" alt="Capture d’écran 2019-10-02 à 12 41 21" src="https://user-images.githubusercontent.com/2655257/66038631-fc811200-e512-11e9-9501-b933de832926.png">

**System information:**
 - P5 Version: 0.6.0 (latest version, installed via pip 19)
 - Python version: Python 3.7.3 (anaconda)
 - Operating system: Mac OS X Mojave

jeremydouglass:
In order to run at double density with a retina display, it seems like (untested) you need to set 

```
builtins.pixel_x_density = 2
builtins.pixel_y_density = 2
```

This would be the equivalent of [displayDensity()](https://processing.org/reference/displayDensity_.html)

Did you try setting these manually? Auto-detection would be better, but the first question is whether density is working at all....

swirly:
Thank you for your investigation

just tried this now
```
from p5 import *
from random import random, randrange
import builtins

builtins.pixel_x_density = 2
builtins.pixel_y_density = 2
```

Does not solve anything. Tried to use 1 or 4. The behaviour does not change. Density seems not to work

swirly:
Any progress on this issue ?

pikeas:
I'm on a Retina MBP and can confirm that this appears to be working correctly with p5 0.7.1.

ziyaointl:
Closed since this is working in 0.7.1.

PR #357, opened 2022-07-14
Enhancing skia further

tushar5526:
### Transform 

- [x] rotate()
- [x] scale()
- [x] shearX()
- [x] shearY()
- [x] translate()
- [x] applyMatrix()
- [x] resetMatrix()

### Settings

- [x] background()
- [x] clear()
- [x] fill()
- [x] noFill()
- [x] noStroke()
- [x] stroke()

### Attributes

- [ ] noSmooth()
- [x] strokeCap()
- [x] strokeJoin()
- [x] strokeWeight()
- [x] ellipseMode()
- [x] rectMode()

### Vertex

- [x] beginContour()
- [x] beginShape()
- [x] bezierVertex()
- [x] curveVertex()
- [x] endContour()
- [x] enshape()
- [x] quadraticVertex()
- [x] vertex()


tushar5526:
Comparison of a few sketches. Skia followed by vispy, results look great and we are getting results similar to web canvas. Skia is really great.

https://user-images.githubusercontent.com/30565750/179371416-1bf55ea1-84a0-4315-b529-89d7ba4453c4.mp4

tushar5526:
Tried #298 
https://user-images.githubusercontent.com/30565750/179372793-492795d2-994a-488a-ad88-7b6de660756b.mp4

p5/core/vertex.py
```diff
@@ -309,8 +355,7 @@ def end_shape(mode=""):
     :type mode: str
 
     """
-    global is_bezier, is_curve, is_quadratic, is_contour
-    assert not is_contour, "begin_contour called without calling end_contour"
```
ziyaointl, line 313: What was the reasoning behind removing this assert?

p5/core/vertex.py
```diff
@@ -210,15 +241,30 @@ def end_contour():
...
+    elif builtins.current_renderer == 'skia':
+
+        vert_data = copy.deepcopy(contour_vertices[0])
+        vert_data[-1]["is_vert"] = contour_vertices[0][-1].get('is_vert', None)
+        vert_data[-1]["move_to"] = False
+
+        contour_vertices.append(vert_data)
+
+        if is_first_contour:
+            vertices.append(vertices[0])
```
ziyaointl, line 263: What was the reaons to append `vertices[0]` for the first contour, but not subsequent contours?

p5/sketch/Skia2DRenderer/renderer2d.py
```diff
@@ -34,12 +56,44 @@ def __init__(self):
         self.typeface = skia.Typeface.MakeDefault()
         self.font.setTypeface(self.typeface)
```
ziyaointl, line 58: nit: remove extra space

p5/sketch/Skia2DRenderer/renderer2d.py
```diff
@@ -109,30 +163,55 @@ def render_image(self, img, *args):
...
+
+    def render(self, fill=True, stroke=True, rewind=True):
         """
         Draw the path on current canvas using paint
         """
         # TODO: Check, do we really have to setColor, and setStyle on each render call
         # TODO: Explore ways to do this, such that it works with pGraphics as well
         if self.style.fill_enabled and fill:
+            self.paint.setStrokeCap(self.style.stroke_cap)
+            self.paint.setStrokeJoin(self.style.stroke_join)
```
ziyaointl, line 197: Do we need to set `StrokeCap` and `StrokJoin` in both `if self.style.fill_enabled` and `if self.style.stroke_enabled`?

p5/sketch/Skia2DRenderer/renderer2d.py
```diff
@@ -291,3 +370,183 @@ def triangle(self, *args):
...
+        if is_curve and shape_kind is None:
+            if num_verts > 3:
+                s = 1 - self.curve_tightness
+                self.path.moveTo(vertices[1][0], vertices[1][1])
+                for i in range(1, num_verts - 2):
+                    b = []
+                    v = vertices[i]
+                    b.append([v[0], v[1]])
+                    b.append([
+                        v[0] + (s * vertices[i + 1][0] - s * vertices[i - 1][0]) / 6,
```
ziyaointl, line 393: Pardon me since this question may be very obvious, but where does the 6 come from in this equation?

ziyaointl (commented):
Extra empty python file?

ziyaointl:
Great work! I left some comments, but majority of them are requests for clarification.

p5/core/transforms.py
```diff
@@ -154,12 +154,21 @@ def scale(sx, sy=None, sz=None):
 
 
 def apply_matrix(transform_matrix):
-    """Apply the given matrix to the sketch's transform matrix..
+    """
+    Apply the given matrix to the sketch's transform matrix.
 
     :param transform_matrix: The new transform matrix.
-    :type transform_matrix: np.ndarray (or a 4×4 list)
+    :type transform_matrix: (3 * 3) if skia is the renderer else (4 * 4) np.ndarray or list
```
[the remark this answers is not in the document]
tushar5526: I will leave a TODO for now, since skia supports [3 * 3 matrix transformations only](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/graphics/skiasharp/transforms/matrix#the-reason-for-the-3-by-3-matrix)

p5/core/color.py
```diff
@@ -199,9 +199,11 @@ def parse_color(*args, color_mode='RGB', normed=False, **kwargs):
 
     return red, green, blue, alpha
 
-
+# NOTE: By default expects a color to be in 255 format, whereas renderer uses color values in normalized forms
```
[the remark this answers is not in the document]
tushar5526: Yes constructor expects color to be within 255 range, then it converts the color into normalized form [0, 1] range. 

I added this note when I had to temporary swap the colour values and ran into a bug. 
```
current_color = p5.renderer.style.color // renderer stores in [0,1] format
new_color = Color(*current_color)  // new_color constructed using values in [0,1] while it needs a [0,255] range values
```

p5/core/vertex.py
tushar5526: This assertion seems a bit wrong. The default value of `is_contour` is `False` so this assertion would raise an error even when no `contour` calls are made. 

Also, `is_contour` whether the current shape has a contour in it. So, we would have to use a different variable to indicate whether `end_contour` was called or not.

p5/core/vertex.py
tushar5526: Close the shape before starting the contour, else we will get a line connecting the contour and the shape.

p5/sketch/Skia2DRenderer/renderer2d.py
tushar5526: Nice catch, yes it should only be set and used in `stroke_enabled` if clause.

p5/sketch/Skia2DRenderer/renderer2d.py
tushar5526: https://github.com/processing/p5.js/blob/9cd186349cdb55c5faf28befff9c0d4a390e02ed/src/core/p5.Renderer2D.js#L767

I was not able to find an explanation but that's how it was done in p5.js

p5/core/transforms.py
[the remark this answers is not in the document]
ziyaointl: Oops I forgot skia is a 2D only renderer. Then yes, supporting only 3*3 matrices make sense. We could probably also improve the documentation to say that 3x3 is expected for 2D renderers and 4x4 is expected for 3D renderers.

p5/core/vertex.py
ziyaointl: Got it! Could you document this in an inline comment?

p5/sketch/Skia2DRenderer/renderer2d.py
ziyaointl: I see - while I think it's ok to use logic from other Processing distributions, let's add links in the code that point to where a particular piece of logic was borrowed from.

p5/core/vertex.py
ziyaointl: I think the default value of `False` actually does not trigger this assert, since `not is_contour` would evaluate to `True`.
I think the original `is_contour` had the function of denoting whether `begin_contour` has been called without `end_contour` being called. Without calling `end_contour`, it doesn't make sense to call `end_shape`, hence the original assert.
Under what condition did this assert trigger for you?

tushar5526: `end_contour` resets the value to `False` after `begin_contour` in the old code. I have removed that in the current code, because we need to know whether `is_contour` was called or not to be passed onto `renderer.end_shape`

parsoyaarihant:
Looking great @tushar5526 . Have we done any performance comparison for this renderer? We can compare the FPS of skia renderer with old renderer and p5js as well. 

I am sure this is great performance improvement. But it will be helpful for users to know these benchmarks.

parsoyaarihant:
Does skia renderer solve Retina display issues? I remember this was a major issue for many Mac users. https://github.com/p5py/p5/issues/131

tushar5526:
> Does skia renderer solve Retina display issues? I remember this was a major issue for many Mac users. #131

Yes, I configured glfw to support retina displays as well. But, I was not able to test it locally because I don't have a MAC with me. Mark has probably tested it locally so I think the pixel Density issues is not there.

tushar5526:
> Looking great @tushar5526 . Have we done any performance comparison for this renderer? We can compare the FPS of skia renderer with old renderer and p5js as well.
> 
> I am sure this is great performance improvement. But it will be helpful for users to know these benchmarks.

I can use the current profiling to show the improvements.

tushar5526:
Hey @ziyaointl I am thinking of starting with the new text and image APIs only after getting this one merged, followed by a release. 

I think it will take some time for me to get the access to  make a release. So I would be glad if you could make the release on pypi

p5/core/vertex.py
ziyaointl: Ah I see. I think the old `is_contour` is still useful as a variable that indicates whether we are in a contour, thus providing us a way to nudge users to not call `end_shape` without calling `end_contour` after every `begin_contour`. I think the use case you've mentioned sounds like that a new variable named `has_contour` might be more appropriate. What do you think?

ziyaointl:
@parsoyaarihant I've been running vispy on a mac retina screen for a while now and it seems to work normally, so the issue probably has been fixed at some point in the past. I haven't been able to test skia on mac though due to a segfault. I'll probably have to install the skia library with debug symbols to find out what is causing that.

@tushar5526 Yes! Let me know when everything else in #360 has been finished, and I'll publish the 0.8.0 tag to pypi.

p5/core/vertex.py
tushar5526: Yes, that sounds right. I have done that. Thanks!

tushar5526:
This PR is good to go. @ziyaointl you can review and merge it :)

p5/core/vertex.py
```diff
@@ -199,7 +230,8 @@ def begin_contour():
     for internal shapes, draw vertices shape in counter-clockwise.
 
     """
-    global is_contour, contour_vertices, contour_vertices_types
+    global has_contour, contour_vertices, contour_vertices_types, has_contour, is_contour
```
ziyaointl, line 233: duplicate has_contour

p5/core/vertex.py
```diff
@@ -210,15 +242,32 @@ def end_contour():
 
     For more info, see :any:`begin_contour`.
     """
-    global is_contour, curr_contour_vertices, curr_contour_vertices_types
+    global has_contour, curr_contour_vertices, curr_contour_vertices_types, is_first_contour, has_contour, is_contour
```
ziyaointl, line 245: same here

p5/core/vertex.py
```diff
@@ -309,46 +358,56 @@ def end_shape(mode=""):
     :type mode: str
 
     """
-    global is_bezier, is_curve, is_quadratic, is_contour
-    assert not is_contour, "begin_contour called without calling end_contour"
+    global is_bezier, is_curve, is_quadratic, has_contour, is_first_contour, has_contour, is_contour
```
ziyaointl, line 361: same here

p5/core/vertex.py
```diff
@@ -83,13 +88,15 @@ def curve_vertex(x, y, z=0):
 
     if p5.mode == "3D":
         return
-    else:
-        if is_contour:
+    if builtins.current_renderer == 'vispy':
+        if has_contour:
```
ziyaointl, line 92: I think this should be `is_contour` instead. If an `end_contour` has been called, we probably shouldn't add more vertices to `curr_contour`.

ziyaointl: (this comment applies to other places in this file as well)

tushar5526: Hi Mark. There was one more thing to this. Once we call a `begin_contour`  and end it with `end_contour` we don't reset the `is_contour` back to `False`. Apparently in p5.js something like 

```
beginShape()
vertex()
...
beginContour()
vertex()
...
endContour()
# vertex called again not in contour
vertex()
...
```
The vertex calls after `end_countour` are just neglected because the `is_contour` flag has been set are added to `contour_vertices`. I am unsure if this is a bug or `beginShape` is intended to be called in a way such that we first draw the shape followed by the contour and then end the shape, I didn't find anything in the docs.

However, to keep our output consistent with p5.js, we would also leave the `is_contour` variable to `True`.

I am now using `in_contour` as a check to see if we are inside a contour or not for the assert.

tushar5526: I have added a comment in the docs for future reference.

ziyaointl: I see, that's a great observation! Thanks for bringing it up!

PR #359, opened 2022-07-25
Upgrading min python version to >= 3.7

tushar5526:
Related to #360 
- Update GHAs to support python >= 3.7
- Bump skia to >= 85.0 and update base.py for the new skia version
- Update setup.py

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -12,13 +12,15 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9']
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python
+      - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v1
         with:
-          python-version: 3.6
+          python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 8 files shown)


tushar5526:
@ziyaointl Should we run the linter and pytest for each python version?

tushar5526:
@ziyaointl I have saved a draft for the next release as well, do take a look at it and share your thoughts.

tushar5526:
We should make the release after merging "enhancing skia" PR.

ziyaointl:
@tushar5526 I think running linter and the formatter using the lowest supported python version (3.7) is ok, since newer versions of Python usually don't break backward compatibility from my experience, but accidentally using a feature only supported in newer versions means we break the library for anyone using lower versions of Python.

ziyaointl:
On publishing to pypi, looks like although I have the permission to release a new version, I do not have the permission to add new maintainers. I will send an email to @abhikpal and @parsoyaarihant on this.

ziyaointl:
nit: I recommend opening multiple PRs for some of the items included in this release (e.g. bumping minimum python version, updating changelog/docs, etc.). This makes changes easy to revert individually if something breaks, and makes the history on master clearer. To track the progress for the release, we can open an issue that includes a checklist.
P.S. tags can be directly pushed to Github without creating a PR
```bash
# Once we have all PRs that we want to include merged to master
git checkout master
git pull
git tag v0.8.0 
git push origin v0.8.0 
```

tushar5526:
I have broken it into multiple PRs

tushar5526:
I will probably update the GHAs to run one after each other |  linters -> pytest -> sanity_tests. If anything fails then we don't have to check further, so we can keep different python version as well.

ziyaointl:
@tushar5526 Yeah that's a good idea!

tushar5526:
I will do it in a different PR this one is good to go.

Issue #360, opened 2022-07-26
New Release v0.8.0


tushar5526:
- [x] Increment the version number in setup.py
- [x] Create a new tag with that version
- [x] Write some update logs summarizing the updates since last update, include this in both the p5py docs and the github release note
- [x] Create a new release on Github
- [x] Create a new release on pypi

tushar5526:
Hey @ziyaointl 

How are we generating the release assets in p5py. I see that in older assets we include the docs as well, whereas with the command `python -m build` docs are not included in the `wheels` or `tar ball`. 

Do we have to generate different builds for GitHub release and pypi release ?

ziyaointl:
I think the assets in previous releases on Github were automatically generated by Github, which were just zipped/tarred source code at the specified tag. Hence the docs were left in. You could continue using the automatically generated assets, or you could also upload the assets generated by `python -m build`. Up to you :)

tushar5526:
Closing this. We have a new release!

PR #365, opened 2022-08-03
Update MANIFEST to include requirements.txt and fix the name of LICENSE file

tushar5526:
The build was failing due to `requirements.txt` not being found. We had to manually include it in the `MANIFEST.in`

Both `tar ball` and `wheels` are building fine now


tushar5526:
Related to #360

ziyaointl (approved):

Issue #371, opened 2022-08-06
Feature: Typography support for Skia


tushar5526:
### Attributes

- [x] [textAlign()](https://p5js.org/reference/#/p5/textAlign)
- [x] [textLeading()](https://p5js.org/reference/#/p5/textLeading)
- [x] [textSize()](https://p5js.org/reference/#/p5/textSize)
- [x] [textStyle()](https://p5js.org/reference/#/p5/textStyle)
- [x] [textWidth()](https://p5js.org/reference/#/p5/textWidth)
- [x] [textAscent()](https://p5js.org/reference/#/p5/textAscent)
- [x] [textDescent()](https://p5js.org/reference/#/p5/textDescent)
- [x] [textWrap()](https://p5js.org/reference/#/p5/textWrap)

### Loading & Displaying

- [x] [loadFont()](https://p5js.org/reference/#/p5/loadFont)
- [x] [text()](https://p5js.org/reference/#/p5/text)
- [x] [textFont()](https://p5js.org/reference/#/p5/textFont)
- [ ] [p5.Font](https://p5js.org/reference/#/p5.Font)

tushar5526:
Hey @ziyaointl, should we implement `preload` for p5py. [Preload docs link](https://p5js.org/reference/#/p5/preload)

> Called directly before [setup()](https://p5js.org/reference/#/p5/setup), the [preload()](https://p5js.org/reference/#/p5/preload) function is used to handle asynchronous loading of external files in a blocking way. If a preload function is defined, [setup()](https://p5js.org/reference/#/p5/setup) will wait until any load calls within have finished

It seems preload waits for the asynchronous functions to finish and then proceeds with `setup` and `draw`.

We can just load stuff before the `setup` in our python sketches so it seems there is no specific need for a preload function IMO.

```
// load fonts, images etc here

def setup()...

def draw().... 
```


Thoughts?

ziyaointl:
I think it's ok to implement `preload` for completeness - it's easy to implement (I don't think we need to do anything extra for async calls) and does not incur much of a performance cost, but I agree that it could be deferred to future releases.

tushar5526:
I have made `implementing preload` a good first issue for folks in the community. It would not affect `loadFont` and other APIs so I can continue with other implementations.

tushar5526:
Turns out `preload` is needed because a different `load_font` will depend upon the `current renderer` and we have to wait till `run()` is called and `p5.renderer` is assigned.

[...]

PR #376, opened 2022-08-08
Implementing Typography for Skia

### LINKED ISSUE #373: Implement preload
Setup a `preload` method for p5py

Reference: [preload](https://p5js.org/reference/#/p5/preload)

tushar5526:
Closes #373 
Related to #371


tushar5526:
Hey @ziyaointl 

I have called the `preload` right there in the `userspace.py` because it seemed the easy way instead of putting it in the `base.py` and calling it somewhere in the `sketch` class.

Also `preload` does not affect `frame_count` or any other sketch settings in any way so I called it in the `userspace.py` as of now.  

Let me know your thoughts, and you can review this PR. It is good to go.

I will do multiple PRs this time and use issue #371 to keep track of the `implementing typography` status, instead of a single PR with many changes.

p5/core/constants.py
```diff
@@ -44,6 +44,7 @@
 RIGHT = "RIGHT"
 TOP = "TOP"
 BOTTOM = "BOTTOM"
+BASELINE = "BASELINE"
```
ziyaointl, line 47: I think it's beneficial to document what these constants are used for, like the other constants in this file.

p5/core/font.py
```diff
@@ -99,7 +88,10 @@ def text(*args, wrap_at=None):
 
     if len(text_string) == 0:
         return
-    return p5.renderer.text(text_string, position, wrap_at)
+    if builtins.current_renderer == "vispy":
+        return p5.renderer.text(text_string, position, wrap_at)
+    elif builtins.current_renderer == "skia":
+        return p5.renderer.text(text_string, *position, wrap_at)
```
ziyaointl, line 94: Is there a reason to make the text API different between renderers? I think we should make them consistent if possible.

p5/core/font.py
```diff
@@ -109,7 +101,7 @@ def text_font(font, size=10):
     :type font: PIL.ImageFont.ImageFont
 
     """
-    p5.renderer.font_family = font
+    p5.renderer.text_font(font, size)
```
ziyaointl, line 104: Instead of passing in a default size of 10, we should probably use the current font size.

p5/sketch/Skia2DRenderer/renderer2d.py
```diff
@@ -29,6 +29,17 @@ class Style2D:
     stroke_cap = skia.Paint.kRound_Cap
     stroke_join = skia.Paint.kMiter_Join
 
+    # Flag that holds whether stroke() is called user
+    stroke_set = False
+    # Flag that holds whether fill() is called user
+    fill_set = False
```
ziyaointl, line 35: This seems to be unused

p5/sketch/Skia2DRenderer/renderer2d.py
```diff
@@ -29,6 +29,17 @@ class Style2D:
     stroke_cap = skia.Paint.kRound_Cap
     stroke_join = skia.Paint.kMiter_Join
 
+    # Flag that holds whether stroke() is called user
+    stroke_set = False
```
ziyaointl, line 33: Can we eliminate the `stroke_set` and always use the default values if the user hasn't set the stroke style?

ziyaointl (commented):
Great work! I really appreciate the effort that you took to split up large PRs into small ones. This makes reviewing so much easier :D

One nit: I think that adding `preload` is conceptually different enough from the other font functions that this PR could be split up further, but this is just for future reference - no need to split up this particular PR now.

ziyaointl:
And yes, I think putting `preload` in `userspace.py` is very reasonable :)

p5/sketch/Skia2DRenderer/renderer2d.py
tushar5526: Texts are drawn a bit differently. By default, they are drawn such that there is no outline or stroke only fill is used. We should not stroke the text until the user has explicitly called `stroke()`, hence the variable `stroke_set` is used. 

Without it, all texts drawn would have an outline as well, which is different from the output of p5.js

p5/core/font.py
tushar5526: I think we can adjust the skia renderer to remove the if-else loops.

p5/core/font.py
tushar5526: Yes, fixed it.

p5/core/constants.py
tushar5526: Updated

tushar5526:
I have updated the PR and added a test for typography

p5/core/font.py
ziyaointl: Thanks! Looks like the `size` parameter is being ignored by vispy's `text_font`, could you also do a quick fix for that as well?

p5/sketch/Skia2DRenderer/renderer2d.py
ziyaointl: I see. I think there's also the `stroke_enabled` variable though. Can we use that to indicate whether the user has called `stroke`? (We'll have to set it to false when user calls `no_stroke`, but I think that's fine).

ziyaointl:
Kudos for adding the tests!

p5/sketch/Skia2DRenderer/renderer2d.py
tushar5526: I am not sure if that would be possible. Things we have to take care of are 
- After each frame these variables are to be reset, so if a sketch ends with `no_stroke()` It should not affect the next draw calls, so we set `stroke_enabled` to `True`
- Text on the other hand is drawn with stroke disabled by default. To get a stroke on the text we have to call stroke() explicitly.
- A single variable would never solve the ambiguity of what must have happened in the sketch for `text` and `shape` APIs. Let's say we have `stroke_enabled` to True by default. Then `text` APIs should somehow know that the current `True` value of `stroke_enabled` is not something we got after explicitly calling `stroke`. 
Similarly, if we would have `stroke_enabled` to `False`, then `shape` APIs would not know whether the `no_stroke` was called or we using the default value. 

So, for that, we would need a `stroke_set` variable.

ziyaointl: Thanks for the explanation! I now understand that we need to variables because `stroke` is default on for shapes and default off for fonts. I do wonder if there are better names for these variables, but this is good for now.

ziyaointl (approved):

PR #380, opened 2022-08-22
Implemented pixel_density for both skia and vispy. Updated the text_wrap in vispy

tushar5526:
Small bug fix and added pixel_density.

### DIFF
--- a/p5/sketch/userspace.py
+++ b/p5/sketch/userspace.py
@@ -48,7 +48,6 @@ __all__ = [
     "save",
     "is_looping",
     "set_frame_rate",
-    # TODO: Add docs for pixel density
     "pixel_density",
 ]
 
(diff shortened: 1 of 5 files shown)


p5/sketch/userspace.py
```diff
@@ -48,6 +48,8 @@
     "save",
     "is_looping",
     "set_frame_rate",
+    # TODO: Add docs for pixel density
```
ziyaointl, line 51: I recommend adding docs along with the functionality in the same PR. It's easier to make sure that docs are in sync with the code. Plus, I don't think writing lots of docs for lots of function in one go is super enjoyable :D

ziyaointl: Aside from this minor point, this PR looks great!

tushar5526: Yes, I agree. I will keep updating the docs as well.

PR #381, opened 2022-08-24
Use framebuffer size to create surface for retina displays

tushar5526:
Always use frame buffer size to create a skia surface to handle retina displays as well, this function will also serve as a utility function for PGraphics object so we need 'size' as well, as window is None in that case.


ziyaointl:
Could you rename the PR so the title is shorter? You can move the extra description needed in the summary. Otherwise this looks good!

PR #382, opened 2022-08-24
Bug fix: image() had wrong args parsing logic

tushar5526:
Refactored image.py to use constants instead of strings

### DIFF
--- a/p5/core/attribs.py
+++ b/p5/core/attribs.py
@@ -194,7 +194,7 @@ def background(*args, **kwargs):
                 no_tint()
                 image_mode("corner")
                 with push_matrix():
-                    image(background_image, (0, 0))
+                    image(background_image, *(0, 0))
 
             return background_image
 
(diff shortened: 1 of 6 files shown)


tushar5526:
Hi @ziyaointl, the mixed-use of tuples and standard variables in function calls is causing trouble here. 
So for a call like `image(img, (x,y), (w, h))`  would lead to `len(args)` being `3` and according to the old logic of args parsing this would lead to the following parsing of args. `img = args[0], and location = args[1:]` which is wrong. 

It seems there is a random mix of using tuples and standard variables. I am thinking of modifying this `image` API only to use standard variables instead of tuples. It does not seem productive to write a complicated parsing logic which supports tuples, standard variables or a random mix of them.

ziyaointl:
Yes, I think decreasing the complexity of the interface is a worthy goal to pursue, and I agree that removing the tuple interface is a good way of doing that. The tuple interface was a good attempt to make p5py more pythonic, but not being compatible with other processing languages at the API level is inconvenient for most existing processing users, and like you said, the complexity of maintaining two APIs is showing its downside of being harder to maintain.

This said, we should be very clear in our communication with the users about this change, since this does break previous p5py programs that use the tuple API. We should make sure that any API deletion occurs alongside a major version bump, and we should highlight the deletion in very obvious ways in the changelog.

Issue #379, opened 2022-08-22
Implement Image APIs


tushar5526:
Implement the following Image APIs

### Image
- [ ] [createImage()](https://p5js.org/reference/#/p5/createImage)
- [ ] [saveCanvas()](https://p5js.org/reference/#/p5/saveCanvas)
- [ ] [saveFrames()](https://p5js.org/reference/#/p5/saveFrames)
- [ ] [p5.Image](https://p5js.org/reference/#/p5.Image)
### Loading & Displaying Pixels
- [ ] [loadImage()](https://p5js.org/reference/#/p5/loadImage)
- [ ] [image()](https://p5js.org/reference/#/p5/image)
- [ ] [tint()](https://p5js.org/reference/#/p5/tint)
- [ ] [noTint()](https://p5js.org/reference/#/p5/noTint)
- [ ] [imageMode()](https://p5js.org/reference/#/p5/imageMode)
- [ ] [pixels](https://p5js.org/reference/#/p5/pixels)
- [ ] [blend()](https://p5js.org/reference/#/p5/blend)
- [ ] [copy()](https://p5js.org/reference/#/p5/copy)
- [ ] [filter()](https://p5js.org/reference/#/p5/filter)
- [ ] [get()](https://p5js.org/reference/#/p5/get)
- [ ] [loadPixels()](https://p5js.org/reference/#/p5/loadPixels)
- [ ] [set()](https://p5js.org/reference/#/p5/set)
- [ ] [updatePixels()](https://p5js.org/reference/#/p5/updatePixels)

tushar5526:
Hey @ziyaointl. It seems we need to have different PImage classes for `skia` and `vispy` renderers, as their implementation is very different. 

We could either declare an abstract class in p5.core and then inherit them in each renderer and do their specific implementation. This doesn't feel necessary to me as we use `p5.renderer` type calls for each type of rendering function, there is no direct need to have an interface which declares the type of `PImage` as renderer will handle those calls. 

So I am thinking of just implementing different renderer specific classes in each renderer. In the same way we did `Style` class.

tushar5526:
Also, `p5.js` uses 1D lists to handle image manipulation as well. So, to specify `RGBA` values, we use `ith`, `i+1th`, `i+2th`, `i+3th` index for each pixel's RGBA values. These manipulations are handled by `load_pixels` and `update_pixels`. 

[...]

PR #383, opened 2022-08-25
Refactored image.py and load_pixels

tushar5526:
Refactored image.py to support image apis for different rendering backends.
`load_pixels` is no longer a context manager as it was different than p5.js and update_pixel is implemented


p5/sketch/Vispy2DRenderer/image.py
```diff
@@ -0,0 +1,473 @@
...
+
+from p5.core import constants
+from p5.core.image import PImage
+
+
+def _ensure_loaded(func):
+    """Reloads the image if required before calling the function."""
+
+    @functools.wraps(func)
+    def rfunc(instance, *args, **kwargs):
```
ziyaointl, line 30: nit: Might be more readable to rename `instance` to `self`, since `self` is more conventional in python.

ziyaointl (approved):

tushar5526:
Related to #379

PR #387, opened 2022-09-07
Add support to draw to an off screen buffer 

### LINKED ISSUE #386: Feature : Support writing to Offscreen buffers in Skia
- [ ] Create a `Graphics` class
- [ ] Implement `create_graphics`

tushar5526:
Related to #386 

The basic design of the Graphics object is like following. 

Graphics is declared as an empty abstract class. In case other renderers, have their own implementations of it. Almost each Processing API is also a method of `Graphics` object so I skipped defining abstract methods for each of them. As that would be too much redundant code and time-consuming. 

In the skia implementation, I am creating a new Renderer, canvas, surface, path and paint for each graphics object.

Then the methods are `binded` to the Graphics objects in a kind of unconventional way, but it works properly. 

The major scope of improvement here is to use a GPU-backed surface for smoother rendering. I need to dig more into how `glfw` contexts work and how can I create an offscreen context for surfaces. 

Currently it's a CPU backed surface and the offscreen sketches are not that smooth. I will open an issue for this.

### DIFF
--- a/docs/reference/graphics.rst
+++ b/docs/reference/graphics.rst
@@ -1,0 +1,20 @@
+Graphics
+========
+
+.. module:: p5
+   :noindex:
+
+create_graphics()
+-----------------
+
+.. autofunction:: p5.core.graphics.create_graphics
+
+
+
+Graphics
+--------
+
+.. autoclass:: p5.core.graphics.Graphics
+   :members:
+   :special-members:
+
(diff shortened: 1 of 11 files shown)


ziyaointl (approved):
This is quite impressive! The `bind` mechanism is simple and elegant. I think not defining a full interface for `Graphics` is fine, we just need documentation/examples somewhere to showcase its general capabilities.

tushar5526:
Yes, I will be updating examples and write a few tutorials for it.

Issue #399, opened 2022-09-24
GL.GL_MULTISAMPLE not supported on Windows and MacOS


ziyaointl:
**Describe the bug**
`GL.glEnable(GL.GL_MULTISAMPLE)` causes p5py to crash under the skia renderer in MacOS and Windows.

**To Reproduce**
Run the any p5 script with skia as renderer.

**Expected behavior**
p5py should not crash

**Screenshots**
N/A

**System information:**
MacOS
 - p5 release (version number or latest commit): 6f0d09632ce67977741f1c2bf3fa68f6d99bb463
 - Python version: Python 3.9.13
 - Operating system: MacOS Montery 12.4

Windows
 - p5 release (version number or latest commit): 6f0d09632ce67977741f1c2bf3fa68f6d99bb463
 - Python version: Python 3.10.7
 - Operating system: Windows 10 Pro 20H2

**Additional context**
Windows stack trace:
```bash
Traceback (most recent call last):
  File "test.py", line 22, in <module>
    run(renderer='skia')
  File "p5\sketch\userspace.py", line 231, in run
    p5.sketch.start()
  File "p5\sketch\Skia2DRenderer\base.py", line 159, in start
    self.window = self.glfw_window()
  File "p5\sketch\Skia2DRenderer\base.py", line 80, in glfw_window
    GL.glEnable(GL.GL_MULTISAMPLE)
  File "OpenGL\platform\baseplatform.py", line 415, in __call__
    return self( *args, **named )
  File "OpenGL\error.py", line 230, in glCheckError
    raise self._errorClass(
OpenGL.error.GLError: GLError(
        err = 1282,
        description = b'invalid operation',
        baseOperation = glEnable,
        cArguments = (GL_MULTISAMPLE,)
)
```
MacOS stack trace:
```bash
$ python -X faulthandler issue352.py
Fatal Python error: Segmentation fault

Current thread 0x0000000111041600 (most recent call first):
  File "OpenGL/platform/baseplatform.py", line 415 in __call__
  File "p5/sketch/Skia2DRenderer/base.py", line 80 in glfw_window
  File "p5/sketch/Skia2DRenderer/base.py", line 159 in start
  File "p5/sketch/userspace.py", line 231 in run
  File "issue352.py", line 40 in <module>
[1]    65875 segmentation fault  python -X faulthandler issue352.py
```

We may need to find another way to enable antialiasing in skia.

tushar5526:
[...]

PR #402, opened 2022-10-11
Use frame buffer value to create initial surface

tushar5526:
Related to #399 
Hey @ziyaointl, I think I was not initializing the frame buffers properly. It should be fixed now. 

Can you test this and also test the behaviour when we resize manually using mouse and dragging the sides of the window on MAC ?

Explanation of Changes: When no size is specified, use the frame_buffer size to create surface.

### DIFF
--- a/p5/sketch/Skia2DRenderer/base.py
+++ b/p5/sketch/Skia2DRenderer/base.py
@@ -108,11 +108,8 @@ class SkiaSketch:
 
     # create a new surface everytime
     def create_surface(self, size=None):
-        if not size:
-            size = self._size
-        self._size = size
-        builtins.width, builtins.height = size
         self.surface = self.skia_surface(self.window, size)
+        builtins.width, builtins.height = glfw.get_window_size(self.window)
         self.canvas = self.surface.getCanvas()
         p5.renderer.initialize_renderer(self.canvas, self.paint, self.path)
 


p5/sketch/Skia2DRenderer/base.py
```diff
@@ -108,11 +108,8 @@ def skia_surface(self, window=None, size=None):
 
     # create a new surface everytime
     def create_surface(self, size=None):
-        if not size:
-            size = self._size
-        self._size = size
```
ziyaointl, line 113: I think self._size still needs to be updated in the new version for this to work:
```
    @property
    def size(self):
        return self._size
```

ziyaointl:
Thanks for taking the time to look into this! I tried this PR out and the behavior is the same as before: the program starts out as a black frame. If I resize the window, the contents do show up but they are smaller than expected (they take about a quarter of the screen).

I wonder if we could do something similar to the vispy renderer in `userspace.py` to get the phyical size of the canvas on the screen and calculate the pixel density: https://github.com/p5py/p5/blob/6b079cf3fb19f691fdd254e3bb3c82c38905cdeb/p5/sketch/userspace.py#L207

Although we can't use the `physical_size` property from vispy directly in skia, it looks like vispy is wrapping around a glfw call anyway: 
https://github.com/vispy/vispy/blob/859fcaedcc54ebe1ecfd21348cefa37a1cf08ca6/vispy/app/backends/_glfw.py#L377

Issue #362, opened 2022-08-01
Update the GitHub Actions workflow


tushar5526:
There is no need to run `sanity_tests` GHA after `linter` or `pytest` fails and vice-versa. 

Update the GHAs to wait for other workflows to finish. 

GHAs should run in this order. 

`Linter GHA` followed by `Pytest GHA` followed by `Sanity_tests GHA`. If anyone of the GHA fails we don't run the workflows after it in the mentioned flow.

tushar5526:
Closed by #412

PR #412, opened 2022-12-29
Fix#362 : Fixed order of GHAs

RyanWalker277:
Related to #362 .

Hey @tushar5526 @ziyaointl ! I have restructured the build tests and compiled all of them into one singe `GHA`.Can you test this?

Explaination of changes: added a `needs` parameter in every job connecting it with the previous job.


.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install pytest
+          sudo apt-get install python-opengl libglfw3 -y
```
tushar5526, line 49: This has to be installed first. See the existing GHAs and keep them the same order. That's why the actions are failing as well.

tushar5526 (changes requested):
GHAs are failing. Install the apt dependencies before hand.

.github/workflows/build-tests.yml
RyanWalker277: The order is the same as existing GHAs.The `python-opengl` isn't installing in the ubuntu image for some reason.I even tried installing the package through a seperate GHA on another repo yet it failed , seems like an issue with the apt-package itself.Can you re-run the pytest action on the latest commit and tell if the same issue persists there as well?

tushar5526: Hey, I just checked can you change it to `python3-opengl` from `python-opengl` and see if it works? I think the name has changed in newer version. https://packages.ubuntu.com/jammy/python/python3-opengl

RyanWalker277: Ya that was the issue.Now it's working fine.

tushar5526 (approved):

tushar5526:
Thank you for the contribution

PR #414, opened 2023-01-10
Check is user has defined preload before use

davidlowryduda:
The current code checks if "setup" is defined by the user before loading "preload". I suspect this is a simple copy-paste mistake from a couple of lines later, where the code checks if "setup" is defined before loading "setup".

This is a tiny fix.


tushar5526:
Thank you for the fix :+1:

Issue #419, opened 2023-02-19
Unexpected behavior of size() in setup()


devAyushDubey:
> This is a more descriptive continuation of #372 

**Describe the bug**
Any `Shape, background()`,etc. when called together with `size()` in `setup()` produces unexpected outcomes. With `background()` and `size()` together in `setup()` the background is not changed, in case of a 2D Primitive like `circle()` the circle is not visible.

**To Reproduce**
Run the following codes:

```python
from p5 import *

def setup():
    size(512,512)
    circle(100,100,50)

run(renderer="skia")  # Also with run()
```

**Expected behavior**
A `512x512` size window with a circle at (100,100).

**Actual behavior**
![Screenshot from 2023-02-19 05-16-47](https://user-images.githubusercontent.com/33064931/219904669-f5b28d31-dbd0-471c-afa4-1d9f95c9407f.png)
Commenting out `size()`:
![Screenshot from 2023-02-19 05-17-25](https://user-images.githubusercontent.com/33064931/219904681-5270ac59-f084-4b59-a8e7-942610df1be5.png)
---
```python
from p5 import *

def setup():
    size(512,512)
    background(255)

run(renderer="skia")  # Also with run()
```
**Expected behavior**
A `512x512` size window with a white(255) background.

**Actual behavior**
![Screenshot from 2023-02-19 05-16-08](https://user-images.githubusercontent.com/33064931/219904765-09ed3759-94c7-4231-bee2-1c9580b915da.png)
Commenting out `size()`:
![Screenshot from 2023-02-19 05-22-29](https://user-images.githubusercontent.com/33064931/219904783-b4ebd6bb-df2c-4df8-85d6-e7574eb65288.png)

**System information:**
 - p5 release: 0.8.2 
 - Python version: 3.10.6
 - Operating system: Ubuntu 22.04.1 LTS (64-bit)

devAyushDubey:
I would like to work on this issue.

tushar5526:
I assume this is something related to setup logic in skia renderer. The swapping of framebuffers is not being done right, I think. You can start looking from there.

Issue #372, opened 2022-08-07
Using background and size in setup would not set the background to the specified color


tushar5526:
Each `size` calls creates a new surface and a new canvas. Any path drawn on the old canvas would be lost. 

```
from p5 import *

def setup():
    size(600,340)
    background(220)

def draw():
    circle(mouse_x, mouse_y, 50)

run(renderer='skia')
```
Background not set to `220` in former sketch

```
from p5 import *

def setup():
    # size(600,340) -> removed size, no new surface and canvas created so it would work as expected
    background(220)

def draw():
    circle(mouse_x, mouse_y, 50)

run(renderer='skia')
```

tushar5526:
Users can for now use the workaround like this, by the time it gets fixed.

```
def draw():
    if frame_count == 1:
       background(220)
    
```

devAyushDubey:
Hey @tushar5526 ,
What I found was, the issue is not just with `size()` and `background()` being together in `setup()` but with any `2D Primitive` when called after size() in setup(), is not visible (probably covered).
The following code will also not work as it is supposed to:
```python
from p5 import *

def setup():
    size(512,512)
    circle(100,100,50)

run()  # Also with run(renderer="skia")
```
Would produce:
![Screenshot from 2023-02-16 16-30-43](https://user-images.githubusercontent.com/33064931/219348335-271bb8de-988e-4a01-be3f-a8616aff50be.png)

Commenting out `size()` gives:
![Screenshot from 2023-02-16 16-32-44](https://user-images.githubusercontent.com/33064931/219348504-99278f49-43f1-44ab-8c82-c8246250df3c.png)

This is quite different from from p5.js , is this something peculiar for p5py and I am missing a reference or there is an issue with `size()` here?

tushar5526:
Yes, as p5py is in its beta version there are a lot of bugs like this to be fixed. You can start an issue and start working on it. For 2D renderers I would prefer that you also try skia renderer and priortize that for fixes related to 2D renderer. 

`run(renderer='skia')`

devAyushDubey:
@tushar5526 really sorry for the late response, was a little occupied. I have created a new issue for the same and would love to work on it. Should I go ahead?

tushar5526:
yes

procub3r:
Hello!
I've investigated this a bit, and when `size()` is called in `setup()`, the reason why `background()` or `ellipse()` or any function that would draw to the screen doesn't work is that:
[...]

PR #420, opened 2023-02-20
Fix for unordinary behavior of size() in setup() for skia

devAyushDubey:
This is a partial "fix" for #419 particularly for Skia Renderer. Related to #420

What I found tracing the calls are following:
- This issue is due to how `GLFW` handles `window resize events`.
- Whenever a glfw window is resized by `size()` or manually by the user, it triggers a window resize event in the `glfw event queue`, this event is inherently asynchronous, so this does not take place instantly, but when it does, a callback function is called. The callback function is assigned by [self.assign_callbacks()](https://github.com/p5py/p5/blob/4c63a4f72785d084f9dffe2ee99fe37a60fe373d/p5/sketch/Skia2DRenderer/base.py#L158) in the `start()` of SkiaSketch.
- We assign [frame_buffer_resize_callback_handler()](https://github.com/p5py/p5/blob/4c63a4f72785d084f9dffe2ee99fe37a60fe373d/p5/sketch/Skia2DRenderer/base.py#L202) as a callback for such events.
- In this callback, for the changed window size, we create a `new surface` with the new dimensions by [self.create_surface(size=(width, height)).](https://github.com/p5py/p5/blob/4c63a4f72785d084f9dffe2ee99fe37a60fe373d/p5/sketch/Skia2DRenderer/base.py#L215)
#### **Now here is the issue:**
  - This newly created surface which is clear and black is the new `self.surface` and is subsequently displayed on the window by the `main_loop()`, this is the reason why anything drawn in `setup()` with `size()` is not visible because a **new surface is created above, everytime is size() is called.**

### Potential Fix:
- Re-render all the contents of `setup()` on the new surface.

This is exactly what I am doing here, I have done 2 changes:
- Called `setup_method()` again after the window is resized and new surface created, so that all the renders of the previous surface is again rendered on the new surface.
- Calling `setup_method()` again, posed the problem of recurrent calling of `size()`, to not let that happen, I have added a simple check in [userspace.py/size()](https://github.com/devAyushDubey/p5/blob/cb18c379df87a4f5c69b5b925e92996752dbd06f/p5/sketch/userspace.py#L257) that ensures that, **`size()` is not called in the re-rendering call**.
- This check together with the added `self.resized = False` in [frame_buffer_resize_callback_handler()](https://github.com/devAyushDubey/p5/blob/cb18c379df87a4f5c69b5b925e92996752dbd06f/p5/sketch/Skia2DRenderer/base.py#L208) also facilitates manual resizing of window and works perfectly fine.

### Now
```python
from p5 import *

def setup():
    size(512,512)
    background(220)
    circle(100,100,50)

run(renderer="skia")
```
Screenshot from 2023-02-20 11-12-19
**With draw()**
Screenshot from 2023-02-20 11-14-27
**Manual Resizing:**
Screencast from 20-02-23 11_16_01 AM IST (1)

### DIFF
--- a/p5/sketch/Skia2DRenderer/base.py
+++ b/p5/sketch/Skia2DRenderer/base.py
@@ -205,7 +205,7 @@ class SkiaSketch:
         Values of width and height may not be equal to the actual window's width and height
         in Retina Display
         """
-
+        self.resized = False
         # Callback handler for frame buffer resize events
         builtins.pixel_x_density = width / self.size[0]
         builtins.pixel_y_density = height / self.size[1]
@@ -213,6 +213,7 @@ class SkiaSketch:
 
         GL.glViewport(0, 0, width, height)
         self.create_surface(size=(width, height))
+        self.setup_method()
         # with self.surface as self.canvas:
         #     # redraw on the canvas/ ( new frame buffer ) after resizing
         #     # and do not rewind/clear the path
--- a/p5/sketch/Skia2DRenderer/base.py
+++ b/p5/sketch/Skia2DRenderer/base.py
@@ -5,6 +5,7 @@ import contextlib, glfw, skia
 from OpenGL import GL
 from time import time
 
+import copy
 from ..events import handler_names
 from .handlers import *
 from .util import *
@@ -214,11 +215,13 @@ class SkiaSketch:
         # To take in account the contents of the last surface.
         old_image = self.surface.makeImageSnapshot()
         old_image = old_image.resize(old_image.width(),old_image.height())
+        old_style = copy.copy(p5.renderer.style)
 
         GL.glViewport(0, 0, width, height)
         self.create_surface(size=(width, height))
         self.setup_method()
         
+        p5.renderer.style = old_style
         # Renders the content of the last surface to the newly created
         with self.surface as self.canvas:
             self.canvas.drawImage(old_image,0,0)
--- a/p5/sketch/Skia2DRenderer/base.py
+++ b/p5/sketch/Skia2DRenderer/base.py
@@ -215,7 +215,7 @@ class SkiaSketch:
         # To take in account the contents of the last surface.
         old_image = self.surface.makeImageSnapshot()
         old_image = old_image.resize(old_image.width(),old_image.height())
-        old_style = copy.copy(p5.renderer.style)
+        old_style = copy.deepcopy(p5.renderer.style)
 
         GL.glViewport(0, 0, width, height)
         self.create_surface(size=(width, height))
--- a/p5/sketch/Skia2DRenderer/base.py
+++ b/p5/sketch/Skia2DRenderer/base.py
@@ -206,30 +206,33 @@ class SkiaSketch:
         Values of width and height may not be equal to the actual window's width and height
         in Retina Display
         """
-        self.resized = False
         # Callback handler for frame buffer resize events
+
+        self.resized = False
         builtins.pixel_x_density = width / self.size[0]
         builtins.pixel_y_density = height / self.size[1]
         self.pixel_density = width * height // (self.size[0] * self.size[1])
 
-        # To take in account the contents of the last surface.
+        # Creates an Image of current surface and a copy of current style configurations
+        # For the purpose of handling setup_method() re-call 
+        # Ref: Issue #419
         old_image = self.surface.makeImageSnapshot()
-        old_image = old_image.resize(old_image.width(),old_image.height())
+        old_image = old_image.resize(old_image.width(), old_image.height())
         old_style = copy.deepcopy(p5.renderer.style)
 
         GL.glViewport(0, 0, width, height)
         self.create_surface(size=(width, height))
         self.setup_method()
-        
-        p5.renderer.style = old_style
-        # Renders the content of the last surface to the newly created
-        with self.surface as self.canvas:
-            self.canvas.drawImage(old_image,0,0)
 
-        #     # redraw on the canvas/ ( new frame buffer ) after resizing
-        #     # and do not rewind/clear the path
+        # Redraws Image on the canvas/ new frame buffer
+        # Previously stored style configurations are restored for
+        # discarding setup_method() style changes
+        p5.renderer.style = old_style
+        with self.surface as self.canvas:
+            self.canvas.drawImage(old_image, 0, 0)
 
         # Tell the program, we have resized the frame buffer
+        # and do not rewind/clear the path
         # Restart the rendering again
         self.resized = True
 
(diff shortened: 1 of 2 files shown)


devAyushDubey:
Let me know your thoughts @tushar5526 🙂

tushar5526:
```
from p5 import *

def setup():
    size(512, 256)
    background(220)
    circle(100, 100, 50)
    stroke_weight(0)

def draw():
    stroke_weight(10)
    circle(mouse_x, mouse_y, 50)
    
run(renderer='skia')

```

Try this example and resize. See the stroke of the circle changes because setup is called again. Also, I am seeing a weird gittery behaviour at my end , some different frame buffer is getting swapped alternatively. Not sure if it is PR specific or different. It would take a proper deep dive though.

procub3r:
> Also, I am seeing a weird gittery behaviour at my end , some different frame buffer is getting swapped alternatively. Not sure if it is PR specific or different. It would take a proper deep dive though.

It's definitely not PR specific. I've noticed this behavior on the current master branch as well.
In fact, this issue prevents my proposal https://github.com/p5py/p5/issues/372#issuecomment-1436111013 from working all the time. It is definitely something worth looking into.

devAyushDubey:
@tushar5526 I think you are right, there was this thing missing in my solution.

Calling `setup_method()` again in [frame_buffer_resize_callback_handler()](https://github.com/p5py/p5/blob/4c63a4f72785d084f9dffe2ee99fe37a60fe373d/p5/sketch/Skia2DRenderer/base.py#L202) would mess up with the `draw()` logic of p5, as during a manual resize the `draw_method()` renders will be overshadowed by the `setup_method()` re-render call.

The problem:
- The problem with the solution was impulsively calling `setup_method()` and re-rendering it's content to the new surface without **taking into account all the renders of the old surface**.

Potential Solution:
- Taking into account the renders on the old surface and **re-rendering all those renders to the new surface** after the `setup_method()` in [frame_buffer_resize_callback_handler()](https://github.com/p5py/p5/blob/4c63a4f72785d084f9dffe2ee99fe37a60fe373d/p5/sketch/Skia2DRenderer/base.py#L202).

I dived a little deep into `GLFW` and `Skia` and found out there is a way we can solve this.
I create an `Image Snapshot` of the old surface by calling [makeImageSnapshot()](https://github.com/devAyushDubey/p5/blob/f08e5b382578e0fc292e861d89b0a0555a5d8ab2/p5/sketch/Skia2DRenderer/base.py#L215) on the old canvas which contains all the renders on that surface, we name it `old_image`.
After the new surface is created and `setup_method()` re-rendered, we render `old_image` on the new surface using [drawImage()](https://github.com/devAyushDubey/p5/blob/f08e5b382578e0fc292e861d89b0a0555a5d8ab2/p5/sketch/Skia2DRenderer/base.py#L224) on the new surface's canvas.

I tried the examples that you mention, they seem to be working perfect on my end, it would be great if you could test it out too.
```python
from p5 import *

def setup():
    size(512,512)
    background(220)
    circle(100,100,50)

def draw():

    circle(mouse_x,mouse_y,50)

run(renderer="skia")
```
Screencast from 21-02-23 10 23 51 AM IST
With `stroke_weight()`:
Screencast from 21-02-23 10 37 26 AM IST

Let me know your thoughts. 🙂

tushar5526:
The problem with calling `setup()` after resizing is that it will change the current style settings of the whole sketch. We need to prevent that from happening. The current solution will still not work. For example, we write a sketch where the background changes to red if we click the left mouse button, and we specify a black background in the setup. On resize, setup settings will be applied. 

I think you are very close to the solution, but instead of calling setup every time frame resizes, we need to figure out a way to copy the existing `verbs` or commands that were sent to canvas and wait for the frame buffers to be resized - and then redraw or execute the commands, followed by continuing with the execution of setup. 

Few other possible solutions could be to preprocess setup function to first execute `size()` wait for frame buffers to adjust and then draw other commands. (Not sure if we can do that in python, but worth a google search)

PS: I don't think we have to be strict with sketch refreshing completely to blank when we resize canvas manually as this happens in p5.js as well.

tushar5526:
> > Also, I am seeing a weird gittery behaviour at my end , some different frame buffer is getting swapped alternatively. Not sure if it is PR specific or different. It would take a proper deep dive though.
> 
> It's definitely not PR specific. I've noticed this behavior on the current master branch as well. In fact, this issue prevents my proposal [#372 (comment)](https://github.com/p5py/p5/issues/372#issuecomment-1436111013) from working all the time. It is definitely something worth looking into.

Yes, not sure if this is due to some glfw/skia updates or code bugs, but this was not the behaviour earlier and maybe this was introduced in some of the recent commits. Feel free to create an issue, I would love to help

devAyushDubey:
I got your point @tushar5526 , we will have to retain our style settings if we are calling `setup()` in the callback.
I have implemented the same, I am using `copy` module to create a `shallow copy` of the style settings before calling `setup()` as [old_style](https://github.com/devAyushDubey/p5/blob/da729cc6befd3558af247d24980e541659d293d5/p5/sketch/Skia2DRenderer/base.py#L218) and assigning this `old_style` as our current style after calling `setup()` in the callback so that all the style changes of `setup()` can be discarded.
With respect to style, it would be like `setup()` was never called during the resize.

```python
from p5 import *

def setup():
    size(512,512)
    background(100)
    no_stroke()
    circle(100,100,50)

def draw():
    circle(mouse_x,mouse_y,50)

def mouse_clicked():
    background('red')
    stroke(0)

run(renderer="skia")
```
Screencast from 23-02-23 12 42 09 PM IST





Also I would be exploring if we can implement some sort of a preprocessing on a `setup()` in python so that we can handle `size()` efficiently in future, for now I think this does the job.

Thank you for being so helpful and patient, do let me know if I am missing out on something 🙂.

procub3r:
Nice fix!

devAyushDubey:
> Nice fix!

Thanks @procub3r 😃

tushar5526:
Looks like a good hack. I think you should use deep copy instead of shallow copy. Otherwise changes in setup() will still be reflected in [mutable](https://note.nkmk.me/en/python-copy-deepcopy/#:~:text=To%20make%20a%20deep%20copy%2C%20use%20the%20deepcopy,function%20of%20the%20copy%20module.&text=In%20a%20deep%20copy%2C%20copies,does%20not%20change%20the%20other.) style properties.

devAyushDubey:
I implemented the required changes. Doing a `deepcopy()` now as it is more logical in our case.

tushar5526:
Hey, I have tested this and this looks good to me. Can you add more verbose comments in the code explaining the logic and also mention the issue number in the callback.

devAyushDubey:
Did the required changes.

tushar5526:
LGTM! Thanks @devAyushDubey

PR #464, opened 2023-12-22
Migrated to poetry

AdityyaX:
### Request for Pull Request Review
#436 
Hey @tushar5526 ,

I've just submitted a pull request  that includes several updates . I'd greatly appreciate it if someone could take some time to review the changes and provide feedback.

#### Changes Made:
- Added Poetry to Project

Your insights and suggestions are invaluable in ensuring the quality and effectiveness of these updates. Your review will help in merging these changes smoothly.

Thank you

PR #470, opened 2024-04-25
Feat/move to poetry

martynvdijke:
Hey all,

Just found this repo while searching for the docs I saw that https://github.com/p5py/p5/pull/464 was still hanging.
So decided to fix that, this introduced poetry as a dependency management tool.
It should also add the pylint configuration as a setting file.


tushar5526:
Thank you for the PR @martynvdijke. It would be great to use [tilde specification](https://python-poetry.org/docs/dependency-specification/#tilde-requirements) to specify the dependency. That way we can still get minor updates for those libraries.

pyproject.toml
```diff
@@ -0,0 +1,43 @@
+[tool.poetry]
+name = "p5"
+version = "0.8.4"
+description = "p5 is a Python library that provides high level drawing functionality to help you quickly create simulations and interactive art using Python"
+authors = ["Your Name <you@example.com>"]
```
tushar5526: You can find the author here - https://github.com/p5py/p5/blob/master/p5/__version__.py#L23

martynvdijke (commented):
Move to use poetry

tushar5526:
Thanks @martynvdijke I will review this soon!

tushar5526:
@martynvdijke thank you for the PR. This looks great to me, I will test this a bit more locally and merge this soon!

martynvdijke:
@tushar5526 any update on this ?

martynvdijke:
Hey hey @tushar5526 I guess you are busy, no worry on that any update on the PR ?

tushar5526:
Hey @martynvdijke ,sorry as I have not been able to find much time (lots of stuff going on rn) I will try to share feedback with you this weekend. Thank you for your patience :)

martynvdijke:
No worries, take your time just checking in :)

martynvdijke:
No worries, take your time just checking in :)

pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+[tool.poetry.dependencies]
+python = "^3.7"
+freetype-py = "^2.1.0"
+glfw = "^2.5.9"
+pillow = "^9.0.1"
+vispy = "^0.10.0"
+pyopengl-accelerate = "^3.1.7"
+pyopengl = "^3.1.6"
+requests = "^2.25.0"
+numpy = [
```
tushar5526, line 17: In the requirements.txt numpy is not tagged to a specific version (which is a bad practice already) 

I appreciate you pinning the version, but why different versions are pinned for different python version?

pyproject.toml
```diff
@@ -0,0 +1,45 @@
+[tool.poetry]
+name = "p5"
+version = "0.8.4"
+description = "p5 is a Python library that provides high level drawing functionality to help you quickly create simulations and interactive art using Python"
+authors = ["Abhik Pal <theabhikpal@gmail.com>"]
+readme = "README.rst"
+
+[tool.poetry.dependencies]
+python = "^3.7"
+freetype-py = "^2.1.0"
```
tushar5526, line 10: Why `freetype-py` is added a direct dependency?

pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+    {version = "*", python = "^3.8"},
+    {version = "^1.21.6", python = "=3.7"},
+]
+dataclasses = [
+    {version = "^0.8", python = "3.6"}
+]
+skia-python = "^87.5"
+
+
+[tool.poetry.group.dev.dependencies]
```
tushar5526, line 27: Why different dependency versions are pinned for different python versions?

tushar5526:
@martynvdijke , again thanks for your PR and patience. I have left some comments. 

I think we can also remove the requirements.txt file if there are no dependency on it.

pyproject.toml
martynvdijke: This is needed because numpy version 1.21 is the last version that is python3.7 compatible.
Otherwise poetry can not figure out the dependency for python 3.7 if that makes sense ?

So by pinning it like this we can ensure compatibility on python3.7 and have a more up to date version for the python versions above.

pyproject.toml
martynvdijke: Removed

pyproject.toml
martynvdijke: Same reason as above otherwise poetry can figure out which dependencies to use for older python versions.
Since they cannot run the newest version on the older pytthon versions

martynvdijke:
@tushar5526 I had a look at the comments and update and explained to your comments

